Issue #7, opened 2017-03-10
Set up CI


adamgundry:
We may or may not have any tests, but we could at least make sure the thing builds on GHC 7.10 and up.

adamgundry:
Resolved by #8.

Issue #6, opened 2017-03-10
Add LICENSE


adamgundry:
The main question is: which one? BSD3?

kosmikus:
BSD3 is the Haskell default. Any reason not to take it? I don't see one.

adamgundry:
Resolved by #8.

Issue #1, opened 2017-03-10
Package up as a proper library


adamgundry:
We should add a `.cabal` file, switch to depending on existing libraries rather than re-defining standard classes, and so on.

The module interface needs some thought. I'd be inclined to go for something like
 * `Optics`: main module that re-exports the most useful bits from other modules
 * `Optics.{Lens,Traversal,...}`: tag definition and functions for working with each different flavour of optic
 * `Optics.Internal`: the newtype and other key things that live inside the abstraction barrier

kosmikus:
You want to use top-level `Optics` rather than `Data.Optics` or `Data.Optic`? I'm ok with that. Does it really make sense to have different modules for each kind of optic?

adamgundry:
I've never really understood why we need a `Data.` (or `Control.`, etc.) prefix on every hierarchical module name, but perhaps there's a good reason I'm missing? I don't feel very strongly about it.

I think one-module-per-optic is useful for documentation purposes, so we can set out the functions typically used with a particular flavour, even if most users actually import the main module.

adamgundry:
Resolved by #8 and #9.

PR #8, opened 2017-03-11
Wip/packaging

kosmikus:
This addresses #1, #6 and #7.

I think we might want multiple packages, with the base package being as dependency-free as possible. I think all these packages should live in the same repo, therefore I created subdirs and created `optics-core` as a package for now (I guess `optics-base` or just `optics` would also work).


kosmikus:
If you want to preserve clean history, I can of course rebase this.

PR #13, opened 2017-03-15
RFC: Add 'Getting' to the hierarchy.

kosmikus:
(I'm in principle ok to merge this for now, but I'm not convinced that this is the right approach, even though it mirrors what the lens package does.)

This is above Fold.

Fold, Traversal, and everything below can be cast to a Getting
with an extra Monoid constraint.

Getter, Lens and everything below can be cast to a Getting without
any extra constraints.

### DIFF
--- a/optics-core/optics-core.cabal
+++ b/optics-core/optics-core.cabal
@@ -16,6 +16,7 @@ library
     Optics.Equality
     Optics.Fold
     Optics.Getter
+    Optics.Getting
     Optics.Iso
     Optics.Lens
     Optics.Optic
@@ -28,6 +29,7 @@ library
     Optics.Internal.Equality
     Optics.Internal.Fold
     Optics.Internal.Getter
+    Optics.Internal.Getting
     Optics.Internal.Iso
     Optics.Internal.Lens
     Optics.Internal.Optic
--- a/optics-core/src/Optics.hs
+++ b/optics-core/src/Optics.hs
@@ -5,6 +5,7 @@ module Optics
 import Optics.Equality  as O
 import Optics.Fold      as O
 import Optics.Getter    as O
+import Optics.Getting   as O
 import Optics.Iso       as O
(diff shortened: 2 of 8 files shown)


kosmikus:
Another option is to keep `Getting`, but only make casting from `Getter` implicit, whereas casting from `Fold` requires an explicit combinator such as:
```
monoid :: (Is k A_Fold, Monoid r) => Optic' k s a -> Getting r s a
```

adamgundry:
Another option I'm wondering about is not making `Getting` part of the hierarchy, but instead providing a combinator to turn a `Fold` into a `Getter`, i.e. something like

    crush :: (Is k A_Fold, Monoid a) => Optic' k s a -> Getter s a
    crush = to . mview

where `mview` is the equivalent of `view` that works for folds. Is that sufficiently expressive?

I'm in two minds about this. On the one hand, the implicit use of `Monoid` when you `view` a fold feels a bit wrong to me. On the other, making `Getting` a proper type in the optic hierarchy matches the `lens` package, and doesn't require a bunch of essentially redundant combinators and corresponding boilerplate.

phadej:
I'm not convinced that this is a good idea. As far as I can tell, the `Getting` *is* `AGetter` using `lens` names (but there aren't `AGetter` neither `AFold`), The problem is similar as with other `A`-optics. Especially `Getting a b` and `Getting b c` don't compose there. 

OTOH in some encoding they might compose. When `Getting a b` and `Getting b c` are folds, then the `Monoid` `mconcat` will happen twice: first all `b`s in `a` will be concatenated into single `b`, and then `c`s in that `b` will be folded into `c`s. That might be different result then single fold of all `c`s in all `b`s of `a`. (I'll draw a picture if that's unclear)

arybczak:
Indeed, I think variants of `view` / `VievableOptic` class solve the core problem. Closing for now.

PR #16, opened 2017-10-12
Hide Constraints type family

adamgundry:
This removes all mention of the `Constraints` type family from the public interface.

Should we also rename the `mk...` functions?


adamgundry:
I've been bold and renamed the constructor functions, plus tidied up `Optics.Internal`.

Issue #18, opened 2018-10-19
Use profunctor encoding internally (and get things going forward)


arybczak:
Hey,

After doing Haskell eXchange presentation about profunctor optics and talking things with @adamgundry and @dcoutts I decided to fiddle more with these to get full hierarchy going (with help of @phadej blog posts, purescript-profunctor-lenses and mezzolens libraries) and see how these realistically behave. I succeeded and here it is:

```haskell

type Optic p a b s t = p a b -> p s t

type Iso a b s t =
  forall p. Profunctor p => Optic p a b s t

type Lens a b s t =
  forall p. Strong p => Optic p a b s t

type Prism a b s t =
  forall p. Choice p => Optic p a b s t

type AffineTraversal a b s t =
  forall p. (Choice p, Strong p) => Optic p a b s t

-- p = Star
type Traversal a b s t =
  forall p. Traversing p => Optic p a b s t

-- p = (->)
type Setter a b s t =
  forall p. Mapping p => Optic p a b s t

-- p = Forget. This is needed because without it 're' would turn 'Prism' into
-- 'Getter', but 'Getter' into 'Review', hence re . re /= id.
type PrismaticGetter a s =
  forall p. Cochoice p => Optic p a a s s

-- p = Forget, view1 for extraction of exactly one value.
type Getter a s =
  forall p. (Bicontravariant p, Cochoice p, Strong p) => Optic p a a s s

-- p = ForgetM, view01 for extraction of at most one value.
type AffineFold a b s t =
  forall p. (Bicontravariant p, Choice p, Cochoice p, Strong p) => Optic p a b s t

-- p = Forget (+ Monoid), viewN for extraction of multiple elements and folding
-- them into one with Monoid instance.
type Fold a b s t =
  forall p. (Bicontravariant p, Cochoice p, Traversing p) => Optic p a b s t

-- p = Tagged. Same concept as PrismaticGetter for re . re = id.
type LensyReview b t
  = forall p. Costrong p => Optic p b b t t

-- p = Tagged, review for reviewing Iso, Prism or LensyReview.
type Review b t =
  forall p. (Bifunctor p, Choice p, Costrong p) => Optic p b b t t
```

![hierarchy](https://user-images.githubusercontent.com/387658/47257712-8fa37200-d491-11e8-9625-5935f98b7cbd.png)

Blue lines mean that optics are convertible between each other with `re`.

[...]

Issue #15, opened 2017-03-20
Affine traversals


adamgundry:
Can we conveniently incorporate affine traversals into the hierarchy? [This article](http://oleg.fi/gists/posts/2017-03-20-affine-traversal.html) might be a useful starting point.

adamgundry:
Implemented in #19.

Issue #10, opened 2017-03-13
Consider using type family instead of functional dependency for Join


adamgundry:

adamgundry:
This would avoid the need for a separate `CanCompose` class.

adamgundry:
Implemented in #19.

Issue #12, opened 2017-03-13
Try to understand indexed optics


adamgundry:

adamgundry:
See #19. Maybe we still need to try and understand indexed optics, but we do at least have them now. :smile:

PR #19, opened 2018-10-23
Use profunctor encoding & other improvements

arybczak:
So this is quite big, but roughly speaking it does the following:

- Converts to profunctor encoding
- Adds affine fold and traversal
- Adds some utility functions for various optics
- Removes vl* functions (apart from vlLens and vlTraversal) as the other ones were using internal versions of typeclasses, so they wouldn't work with external ones
- Reorganizes subtyping module for better error messages and no orphan instances

Implements #18, solves #10 and #15.


arybczak:
I fixed compilation with GHC 8.2.2 (it looks like some weird bug, it was typechecking with 8.0.2 and 8.4.3.

adamgundry:
Thanks for this! There's a lot here to page back in and review, so apologies that it is taking us a while to find time to dig into it. I'll try to go through the details and understand the impact of these changes.

One question is to what extent this unavoidably moves us away from API-compatibility with `lens` (although we haven't really verified to what extent that held in practice).

arybczak:
> One question is to what extent this unavoidably moves us away from API-compatibility with `lens` (although we haven't really verified to what extent that held in practice).

The only thing I can think of is that there is no `from` to turn `Iso` around, as `re` can be more general with profunctor encoding, thus it's also used for Iso.

Other than that, what I noticed:
- We can't use `traverse` directly, but I added a traversal called `traversed` mirroring similar helpers for other optics such as `mapped` or `folded`.
- `.` is `%`.
- `view` is now more compatible with `lens`, but it uses a type class and chooses `view1`, `view01` or `viewN` depending on the optic it's called with.

What is more (this is not related to the encoding, rather than wrapping things in newtypes), optic transformers (such as e.g. [choosing](https://hackage.haskell.org/package/lens-4.17/docs/Control-Lens-Lens.html#v:choosing)) cannot be written for many flavors as one function (as an example, see my implementation of `re` - there is one internal implementation, but it needs to be done with the help of typeclass where a flavor of optic is known, otherwise GHC can't solve the constraints. There isn't many functions like that so I guess it's not a big deal, but I'm just pointing it out.

Actually, if it comes to `choosing`, I don't think it can be written as easily for profunctor encoding as in `lens` - see my question [here](https://stackoverflow.com/questions/52944118/problem-with-defining-choosing-function-for-a-profunctor-lens). I've since then concluded that it's not possible to use `SumProfunctor` to do that because then I couldn't write appropriate `Re` instance - I've gotten further than in the post, but at the end it boils down to the fact that you can split the merged optics, but you can't decide which path you are supposed to take as that depends on the input value and there is no access to it. But still, if it had a single implementation, it would have to be done with the type class anyway (just as `re`), so that's not that big of a deal - we can write `choosing` for all relevant optic types, it will just be different for each of them (without introducing any other profunctor constraints).

arybczak:
I added implementation of indexed traversals, folds and setters. Maybe stacking all of it in a single PR is not the best idea, but in principle profunctor encoding doesn't bring anything new, so I wanted to make it more complete and appealing to review (heh).

Implements #12.

This explores idea(s) from http://oleg.fi/gists/posts/2017-04-26-indexed-poptics.html (specifically, CPS variant) to make regular optics "index-preserving" (which is reflected by the additional type parameter), so that both indexed and regular optics have the same form and can seamlessly compose.

```haskell
λ> iover (_Left % imapped % _1) (,) $ Left [('a', "hey"), ('b', "there"), ('c', "bye")]
Left [((0,'a'),"hey"),((1,'b'),"there"),((2,'c'),"bye")]
```

The difference from how lens library handles these:

- All indices from indexed optics carry through with the usual composition (lens only preserves the rightmost index by default)

lens:
```haskell
λ> iover (itraversed . itraversed) (,) ["hey", "bye"]
[[(0,'h'),(1,'e'),(2,'y')],[(0,'b'),(1,'y'),(2,'e')]]
```

optics:
```haskell
λ> iover (icompose (,) % itraversed % itraversed) (,) ["hey", "bye"]
[[((0,0),'h'),((0,1),'e'),((0,2),'y')],[((1,0),'b'),((1,1),'y'),((1,2),'e')]]
```

- Indexed optics can't be used as regular ones. This is a conscious API decision as internal implementation allows for that (i.e. in principle `IxTraversal` can be implemented via `Traversal` etc.), but without the `conjoined` trick using indexed optics as regular ones incurs a performance loss as opposed to using equivalent, regular optic directly, so I opted out from doing it this way (I'm also not a fan of implicit degradation like that).

adamgundry:
Interesting! Thanks for exploring this. I've had a brief play around with it, and I quite like this approach to indexed optics. Here are some scattered initial thoughts...

Have you thought about coindexed prisms at all?

The Haddocks should explain the `i` and `o` parameters of `Optic`, perhaps with a brief comment and a link to `Optics.IxTraversal` with a discussion of how to understand and compose indexed optics.

The reorganisation means that the definitions of optic flavours are now a bit less modular, although we couldn't add in optic flavours to the subtyping hierarchy after the fact anyway, so this is no great loss. I imagine that making `Join` a closed type family is a bit more flexible in other ways.

In #18 you sketched a hierarchy in which `re` was always reversible, but it seems you haven't done that here. Did you try it? Or do you think the extra flavours are not worth it?

The `CheckIndices i o` overlapping instance breaks inference, I think. For example `itoListOf itraversed "hello"` fails due to overlap, but works without this instance. Perhaps using a closed type family rather than overlapping instances might behave better? More generally I find it a bit puzzling that `CheckIndices` doesn't carry any evidence.

Generalising `view` using a type class seems plausible to me, as it makes the differences for different optic flavours a bit more explicit than in `lens`. I like the fact that things like `:t view @A_Lens` give sensible answers.

arybczak:
> Have you thought about coindexed prisms at all?

Not really, I was deterred by 9 type variables ;)

> In #18 you sketched a hierarchy in which `re` was always reversible, but it seems you haven't done that here. Did you try it? Or do you think the extra flavours are not worth it?

I had it done without the `Optic` newtype. I pushed a commit with them as a test. I though they might not be of much use as in practice no one will probably call `re` twice on an optic to get an original, but I suppose it's good to include it for elegancy's sake. Here's the updated hierarchy (I also added `Equality` for good measure):

hierarchy
 
> The `CheckIndices i o` overlapping instance breaks inference, I think. For example `itoListOf itraversed "hello"` fails due to overlap, but works without this instance. Perhaps using a closed type family rather than overlapping instances might behave better? More generally I find it a bit puzzling that `CheckIndices` doesn't carry any evidence.

Right, I fixed that. `CheckIndices` is a bit tricky. It needs to be a class rather than type family because we need the `o ~ (i -> i)` equality to always be there for unsolved constraint in order for the definition of `itoListOf` etc. to typecheck. On the other hand we can't include it directly in the function signature, because then GHC picks it up and for unflattened indices immediately returns `cannot construct the infinite type` error. However, when the equality is on the LHS of the instance head, it picks the appropriate instance first, then checks the LHS, sees bogus type equality leading to `infinite type` error, but it also sees `TypeError` and then it ignores the bogus equality and returns custom type error instead. Though I wonder if it works by accident.

> Generalising `view` using a type class seems plausible to me, as it makes the differences for different optic flavours a bit more explicit than in `lens`. I like the fact that things like `:t view @A_Lens` give sensible answers.

Hmm, yes, I didn't think of its usage like that. Now I feel less bad for introducing these classes, haha.

adamgundry:
> > Have you thought about coindexed prisms at all?
> 
> Not really, I was deterred by 9 type variables ;)

Heh, I don't blame you... I do wonder if it is possible to get away with a single type parameter that gets instantiated to a promoted datatype with multiple fields, i.e. something like
```hs
data IX = I Type Type Type Type -- i j k l
newtype Optic k (ix :: IX) s t a b
```
but perhaps it's difficult to keep the parameter as a variable even when not using (co)indexed stuff.

Thanks, I now understand `CheckIndices` a bit better. I'd missed the superclass constraint carrying the evidence. It does seem a bit dubious to rely on GHC giving the right error message here, but if it turns out to be unreliable one can probably use an equality datatype instead.

adamgundry:
We agreed to merge this and continue polishing on `master`. I've raised issues for the outstanding things of which I'm aware.

PR #33, opened 2019-01-08
Port a few modules from lens in preparation for TH optics generation

arybczak:
Summary:
- Port `Data.Set.Optics`
- Port `Data.Tuple.Optics`
- Port `Language.Haskell.TH.Optics`
- Port `Optics.Each`
- Add `viewM` as a generalization of `view` (this is regular `view` from lens)
- Add more instances of `*WithIndex` type classes

A few more modules still needed to bring TH optics generation.


arybczak:
Hmm, I edited `doctest` in Makefile and yet it still fails. @phadej is that comand defined somewhere else?

phadej:
`Makefile` is not related to `.travis.yml`: regenerated the latter.

optics-core/optics-core.cabal
```diff
@@ -62,12 +70,21 @@ library
...
+                      DeriveFunctor
                       GADTs
+                      LambdaCase
+                      MultiParamTypeClasses
                       RankNTypes
                       ScopedTypeVariables
                       TupleSections
                       TypeFamilies
+                      TypeOperators
+                      UndecidableInstances
```
phadej: I don't feel nice of having this in `default-extensions`.

optics-core/optics-core.cabal
```diff
@@ -11,15 +11,23 @@ library
   hs-source-dirs: src
 
   build-depends: base >= 4.9 && < 5
+               , array >= 0.5.1.1 && < 0.6
+               , containers >= 0.5.8.1 && < 0.7
```
phadej: Based on https://ghc.haskell.org/trac/ghc/wiki/Commentary/Libraries/VersionHistory we probably want `>= 0.5.7.1`

optics-core/optics-core.cabal
```diff
@@ -11,15 +11,23 @@ library
...
+               , mtl >= 2.2.1 && < 2.3
```
phadej: yet here we can have `>= 0.2.2`

optics-core/src/Data/Tuple/Optics.hs
```diff
@@ -1,47 +1,1280 @@
...
+-- Stability   :  provisional
+-- Portability :  Rank2Types
+--
+-------------------------------------------------------------------------------
 module Data.Tuple.Optics
-  ( _1
-  , _2
-  , curried
-  , uncurried
-  , swapped
```
phadej: these three, `curried`, `uncurried` and `swapped` are moved to `Iso`, or are they gone?

optics-core/optics-core.cabal
phadej: I see, we need this for `Seq.foldlWithIndex`, thus we can either narrow the GHC versions or add a shim.

optics-core/optics-core.cabal
adamgundry: Personally I'm not a big fan of `default-extensions` at all, although it's less of a big deal for "simpler" extensions. At the very least, I'd agree with not enabling `UndecidableInstances` everywhere.

phadej:
@arybczak if you don't mind, I'll move UndecidableInstance to file(s) and merge.

arybczak:
@phadej Sure, please do.

optics-core/optics-core.cabal
adamgundry: > I'm fine with using UndecidableInstances on module basis, but can you elaborate why it's potentially a bad idea? After all, all it does is potentially allows instance resolution to loop.

Well, precisely because it allows instance resolution to loop. :wink: I don't think it is something to be avoided at all costs, just that one should make a conscious decision to use it for particular instances rather than having it globally enabled and hence potentially delaying finding buggy instances. Ideally GHC would even offer a per-instance pragma, but for now per-module is the closest we can get.

PR #36, opened 2019-01-08
Docs structure

phadej:


optics-core/src/Optics.hs
```diff
@@ -1,29 +1,79 @@
+-- |
+--
+-- Module: Optics
+-- Description: The main module, usually you only need to import this one.
```
phadej: description is visible in the module listing. Highlighting this one will make people click it. Hopefully. (Hackage and Stackage may render the module listing differently though).

optics-core/src/Optics.hs
```diff
@@ -1,29 +1,79 @@
...
+--
+-- Introduction...
+--
+-- TODO: motivation behind @optics@
```
phadej: Is the `Talk.pdf` the best source to fill these gaps?

PR #38, opened 2019-01-08
Codegen Is relation

phadej:
I'm not always adding new optic kinds... but when I do, I'd like a machine to help me.

`Topograph.hs` is an old work of mine, which I hadn't yet wrapped into a proper library.


phadej:
... and I'd like to extend this to generation of `Join` if this looks fine otherwise.

PR #39, opened 2019-01-09
Add diagram

phadej:


adamgundry:
As you might have guessed, I like this style of documenting optic flavours such as lens with formation/introduction/elimination rules. I sometimes wonder if those terms assume too much type-theoretic background though?

Issue #41, opened 2019-01-10
Make an Optic interface a little more abstract (hide indexed implementation details)


phadej:
Problem: Talk.pdf says "This just works, because our Lens type isn’t polymorphic"

But it's not true. The true 'Lens' type in current @optics@ is

    forall i. Lens i s t a b

i.e. we __have__ leave it free. Not good.
(We can set it to () or Void, to put into container,
 but then we cannot compose with those optics)

Also the double indices __leak__ the implementation details. Not good either.

Solition: We can cure this, with some additional hackery.
And we can make CheckIndices less hacky, as it would only match on type level list!
I only comment the parts which are changed.

The idea is to use 

    newtype Optic (k :: OptikKind) (is :: [Type) s t a b
        Optic (forall p i. Constraints k p => p i a b -> p (Curry is i) s t)

abstract interface. I.e. move the first (of double indices) to be inside the
Optic.

Note: The implementation is still a compromise.
- A_Setter has always empty index list
- An_IxSettter has always non empty index list
But we are forced to it this way so Constraints reduce properly.  i.e. we can
`sub` Setter into IxSetter, but it would be unusable (as it would have empty
index list.

We can use CheckIndices (and CheckNonIndices if we like) to get better errors.
I think latter is not needed as library doesn't give means to construct
non-indexed optic with non-empty index-list.

More elegantly we'd only have `A_Setter` and differentiate
between indexed and non-indexed variants based on index-list.
But then the type-checker will fight back.

- Making Constraints (k :: OpticKind) (is :: [*]) p breaks `Is` implementation:
  in implies `is` is rigid, so type-family cannot match on it (or we need more
  proofs)

- Using `Constraints An_Traversal p = IxTraversing p`
  but then e.g. `over` would need to be implemented using `IxFunArrow`,
  e.g. passing some dummy index value. Not good.

The full gist is in https://gist.github.com/phadej/0a1b1aa95f102a78d4e26b82f45da903

@arybczak What do you think about this?

arybczak:
Do you think it can be made to play nicely with stuff in #44? Also, why can't we precompose icompose* functions with this version?

adamgundry:
Interesting. I rather like the fact that indexing becomes actually orthogonal to optic kind, and that we reduce the number of points in the lattice.

[...]

PR #43, opened 2019-01-10
Optics.Re and documentation

phadej:


optics-core/src/Optics.hs
```diff
@@ -109,15 +109,83 @@ module Optics
...
+  -- 'x'
+  --
+  -- /Note:/ there are no @from@ combinator.
 
-  -- ** Re
+  , module Optics.Re
 
-  -- | TODO
---  , module Optics.Re
+  -- ** Indexed optics
```
adamgundry: I guess this section may need a bit of tweaking if we do #41.

optics-core/src/Optics.hs
```diff
@@ -109,15 +109,83 @@ module Optics
...
 
-  -- ** Indexed optics
+  -- ** Re
 
-  -- | TODO
-  , module Optics.Indexed
+  -- | Some optics can be reversed with 're':
+  -- @'Iso' i s t a b@ into @'Iso' i b a t s@,
+  -- @'Getter' s t a b@ into @'Review' i b a t s@ etc.
+  -- Red arrows illustrate how 're' transforms optics:
```
adamgundry: This is nice, thanks. The "Core definitions" section above still mentions `re`, even though the diagram no longer includes the red arrows.

optics-core/src/Optics.hs
phadej: True. I'll plan to decide (and implement) #41 first.

optics-core/src/Optics.hs
```diff
@@ -109,15 +109,83 @@ module Optics
...
+  -- >>> let _Identity = iso runIdentity Identity
+  -- >>> view1 (_1 % re _Identity) ('x', "yz")
+  -- Identity 'x'
+  --
+  -- Yet we can use a 'Lens' as a 'Review' too:
+  --
+  -- >>> review (re _1) ('x', "yz")
+  -- 'x'
+  --
+  -- /Note:/ there are no @from@ combinator.
```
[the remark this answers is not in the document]
phadej: Made an issue to discuss that

optics-core/src/Optics.hs
phadej: I made some corrections, but need to think properly how to approach indexed optics in the docs.

Issue #25, opened 2018-12-13
Operations to convert indexed optics to their unindexed equivalents


adamgundry:
It should be possible to turn an indexed traversal into a regular traversal, for example. Our current plan is to make this an explicit operation rather than have it happen implicitly. See also https://github.com/well-typed/optics/pull/19#issuecomment-442521800.

arybczak:
Resolved via #44.

PR #44, opened 2019-01-11
Support for explicit downcasting indexed optics to regular ones

arybczak:
Proof of concept for #25 for now (in particular I think we should have a class with associated function `unIx` instead of `unIxTraversal` etc.

I think it can look better when representing indices as in #41, then regular versions have to accept optics with any list of indices instead of two free type parameters.

@phadej thoughts?

### DIFF
--- a/optics-core/src/Optics/Internal/Optic/Types.hs
+++ b/optics-core/src/Optics/Internal/Optic/Types.hs
@@ -86,12 +86,12 @@ type family Constraints (k :: OpticKind) (p :: * -> * -> * -> *) :: Constraint w
   Constraints 'A_Prism            p = Choice p
   Constraints 'A_PrismaticGetter  p = Cochoice p
   Constraints 'An_AffineTraversal p = (Strong p, Choice p)
-  Constraints 'A_Traversal        p = Traversing p
+  Constraints 'A_Traversal        p = TraversingWithIndex p
   Constraints 'An_IxTraversal     p = TraversingWithIndex p
-  Constraints 'A_Setter           p = Mapping p
+  Constraints 'A_Setter           p = MappingWithIndex p
   Constraints 'An_IxSetter        p = MappingWithIndex p
   Constraints 'A_Getter           p = (Bicontravariant p, Cochoice p, Strong p)
   Constraints 'An_AffineFold      p = (Bicontravariant p, Cochoice p, Strong p, Choice p)
-  Constraints 'A_Fold             p = (Bicontravariant p, Cochoice p, Traversing p)
+  Constraints 'A_Fold             p = (Bicontravariant p, Cochoice p, TraversingWithIndex p)
   Constraints 'An_IxFold          p = (Bicontravariant p, Cochoice p, TraversingWithIndex p)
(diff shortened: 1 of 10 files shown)


optics-core/src/Optics/Indexed.hs
```diff
@@ -1,13 +1,13 @@
 module Optics.Indexed
   (
   -- * Functors with index
-    FunctorWithIndex (..)
+    IxFunctpr (..)
```
phadej: Functpr

Is renaming to avoid clashes with `lens`?

optics-core/src/Optics/Internal/Optic/Types.hs
```diff
@@ -86,12 +86,12 @@ type family Constraints (k :: OpticKind) (p :: * -> * -> * -> *) :: Constraint w
   Constraints 'A_Prism            p = Choice p
   Constraints 'A_PrismaticGetter  p = Cochoice p
   Constraints 'An_AffineTraversal p = (Strong p, Choice p)
-  Constraints 'A_Traversal        p = Traversing p
-  Constraints 'An_IxTraversal     p = TraversingWithIndex p
-  Constraints 'A_Setter           p = Mapping p
-  Constraints 'An_IxSetter        p = p ~ IxFunArrow
+  Constraints 'A_Traversal        p = IxTraversing p
+  Constraints 'An_IxTraversal     p = IxTraversing p
```
phadej: I feel uneasy about this.

Do we still have

```haskell
traverseOf traversed = traverse
traverseOf (traversed % traversed) = traverse . traverse
```

(I think `inspection-testing`)

But if this works, then #41 can be tweaked even further into nicer form indeed.

phadej:
@arybczak could you pull the "Ranem indexed classes" (fixing the typo) into own PR so we could merge it right away (If that's something we want to do, I have no strong opinion about)

optics-core/src/Optics/Indexed.hs
arybczak: Not really, I just figured that since we have IxProfunctor and indexed optics start with Ix, classes can be the same for consistency.

optics-core/src/Optics/Internal/Optic/Types.hs
arybczak: Yeah, nothing changes with regard to operations on regular optics, the extra constraint is just so that indexed variants can be used as regular ones (as Ix* instances for Forget, Star and FunArrow simply ignore indices).

arybczak:
Right, I'll do that later today.

Yes, we want explicit (not implicit though as in lens) downcasting.

optics-core/src/Optics/Internal/Optic/Types.hs
arybczak: Ah, you mean code generation? I didn't check.

phadej:
@arybczak I'll push inspection tests very soon.

arybczak:
Looks like they pass. I force pushed the UnindexableOptic class and removed the commit that renames the classes.

Issue #48, opened 2019-01-11
Rename *WithIndex to Ix*


phadej:
This is good for consistency, and as a bonus avoids clashes with `lens`.

I'll do this when there are no much other stuff in flight.

adamgundry:
I'm not sure about this. Morally the classes are *the same* as the ones in `lens`, aren't they? (Although it seems `lens` includes optics in the class dictionaries, presumably for performance reasons.) So in some sense our versions are really replacements for the `lens` ones and using the same names might make their relationship clearer.

phadej:
A compromise is to have `FunctorWithIndex`, `FoldableWithIndex`, `TraversableWithIndex`, but name profunctors with `Ix` prefix.

Or then all `*WithIndex`  (we have `IxProfunctor` now).

arybczak:
Having internal classes with Ix prefix sounds fine to me, public ones can stay as-is so that people used to lens naming aren't confused.

arybczak:
Resolved via #49.

PR #49, opened 2019-01-11
Rename internal profunctor classes related to indexed optics

arybczak:
Solves #48.


arybczak:
Ok, something broke horribly with this PR, it sometimes is visible and sometimes is not and I can't delete it.

PR #51, opened 2019-01-12
Add PermeableOptic + Operators.State

phadej:


phadej:
I'd rather use ticked versions everywhere. IMHO it's better to embrace the `DataKinds`.

PR #52, opened 2019-01-13
Remove OpticKind and use empty data types for optic tags for clarity

arybczak:


phadej:
I cannot decide
- The fact that even error messages are unticked is great.
- OTOH that we ened to not export `OpticKind` constructors feels limiting.

adamgundry:
Can we export `OpticKind` with its constructors internally (possibly including in an "Internal" but exposed module), but leave out the constructors from the exports of the main public modules?

phadej:
Yes, let's export them from internal module. If we do so, we get:
```
λ *Optics> :t (_1 % _1)
(_1 % _1)
  :: (Field1 s t u v, Field1 u v a b) =>
     Optic 'Optics.Internal.Optic.Types.A_Lens o o s t a b
```

---

*otherwise* we get:

```
λ *Optics> :t (_1 % _1)
(_1 % _1)
  :: (Field1 s t u v, Field1 u v a b) =>
     Optic 'Optics.Internal.Optic.Types.A_Lens o o s t a b
```

i.e. reference to something without any documentation (bad).

kosmikus:
I don't understand the discussion here.

phadej:
@kosmikus I guess, we debate whether we should have `data OpticKind` at all or have (revert to) `data An_Iso, data A_Lens` i.e. not use `DataKinds`.

kosmikus:
Ok, if that's the discussion, can we get all arguments either way listed before making the final decision?

arybczak:
Alright, I pushed an additional commit that reverts usage of OpticKind for comparison's sake.

**OpticKind**

Pros:
- No way to get something that doesn't make sense, i.e. `Optic Int` (though notably there is no way to end up in such situation by accident)
- Somewhat more elegant

Cons:
- Either we don't have type synonyms for tags without backticks and we get backticks everywhere, including error messages:
```
λ> :t mapped % to snd

<interactive>:1:1: error:
    • 'A_Setter cannot be composed with 'A_Getter
    • In the expression: mapped % to snd
```
or we have type synonyms and then we either
* have two ways to annotate an optic (using promoted value or its corresponding type synonym) which creates confusion, or
* jump through some hoops and not export backticked versions or don't export them in the main `Optics` module, but then we get this (with #65):

```
λ> :t mapped % mapped
mapped % mapped
  :: (Functor f1, Functor f2) =>
     Optic
       'Optics.Internal.OpticKind.A_Setter o o (f1 (f2 a)) (f1 (f2 b)) a b
```
whereas ideally we would want this:
```
λ> :t mapped % mapped
mapped % mapped
  :: (Functor f1, Functor f2) =>
     Optic A_Setter o o (f1 (f2 a)) (f1 (f2 b)) a b
```
(which is what we get with this PR).

**Empty data types**

Pros:
- No confusion about backticks
- Clearer type signatures of composed optics and no backticks in error messages

Cons:
- You can get ill-formed things such as `Optic Int` (you need to really try though)

For me empty data types are a clear winner.

kosmikus:
@arybczak Thanks!

I don't think I like the risk of having module-annotated optic names with ticks. While I don't love them, I also don't hate the ticks in error messages as much, so as long as we can prevent them from appearing in qualified form, that'd still be ok with me.

However, I agree that the benefit of having a data kind is really small.

We could make it even smaller by doing something like
```
data OpticKind_
type OpticKind = OpticKind_ -> Type

data A_Lens :: OpticKind
...
```

Then it's also impossible to put anything accidentally wrong in an argument position of `Optic`.

kosmikus:
I don't think so. You cannot match on a type family. We need to match on optic flavours in order to define `Join`.

adamgundry:
The backticks and need for `DataKinds` do make the case for a promoted data type less compelling, I agree. @kosmikus' trick using `OpenKind_` might be a good compromise, assuming it doesn't lead to confusing Haddocks/inferred types/errors?

phadej:
I pushed a commit: https://github.com/well-typed/optics/commit/46be2650f8870040cdfaeab74b274f85ade93258

`TypeInType` is required to be able to use `OpticKind` as a kind.

Inferred types are good

```haskell
λ *Optics> :t mapped % mapped
mapped % mapped
  :: (Functor f1, Functor f2) =>
     Optic A_Setter o o (f1 (f2 a)) (f1 (f2 b)) a b
```

arybczak:
You don't need `TypeInType` if `type OpticKind = OpticKind -> *`. Also, there's still this:

```
λ> :kind! Join A_Lens A_Prism
Join A_Lens A_Prism :: Optics.Internal.Optic.Types.OpticKind_ -> *
= An_AffineTraversal
```

phadej:
```
GHCi, version 8.4.4: http://www.haskell.org/ghc/  :? for help
Loaded package environment from /home/ogre/.ghc/x86_64-linux-8.4.4/environments/default
Loaded GHCi configuration from /home/ogre/.ghci
λ> data OpticKind_
λ> :m +Data.Kind
λ Data.Kind> type OpticKind = OpticKind_ -> Type
λ Data.Kind> data Foo :: OpticKind

<interactive>:4:13: error:
    • Type constructor ‘OpticKind’ cannot be used here
        (Perhaps you intended to use DataKinds)
    • In the kind ‘OpticKind’
λ Data.Kind> :set -XDataKinds
λ Data.Kind> data Foo :: OpticKind

<interactive>:6:13: error:
    • Type constructor ‘OpticKind’ cannot be used here
        (Perhaps you intended to use TypeInType)
    • In the kind ‘OpticKind’
λ Data.Kind>
```

arybczak:
Yes, that's why you need to replace `Type` by `*`.

arybczak:
Ah right, I've had project configured for compilation with GHC 8.6.2, my bad. But the point about somewhat weird result of `:kind!` remains.

kosmikus:
Yes, type synonyms used as kinds require `TypeInType` prior to ghc-8.6. Not sure if that should scare us.

kosmikus:
I'm neutral about the kind annotation of the result of `Join`. Personally, I find it mildly confusing if it's `*`. Others might much prefer it to be. But we should possibly just keep it at `*` for now and change it again once GHC supports unpromoted (and therefore un-ticked) data kinds.

adamgundry:
It sounds like the old-fashioned empty data types are the least worst option for now.
 * No stupid ticks anywhere.
 * No requirement for `DataKinds` in user code.
 * Public interface doesn't need to export multiple copies of the optic kinds.
 * Ill-kinded things are possible but unlikely to be a big problem in practice.

Perhaps for documentation purposes we could have `type OpticKind = Type` and use the synonym if GHC lets us and it doesn't impose additional extension requirements on user code? I'm fine with requiring `TypeInType` in order to define optic kinds, just not to use them.

We can potentially revisit this in the future, especially if GHC gets support for data kinds.

phadej:
I think that if the top priority is to avoid ticked types, then using `*` is less hacky. `Join` has catch all 

```haskell
  -- Everything else is a type error.
  Join k l = TypeError ('ShowType k
                        ':<>: 'Text " cannot be composed with "
                        ':<>: 'ShowType l)
```

also after #59 is merged, the second argument is of kind `[Type]` which will introduce a little of "kindness" into `Optic`, i.e. `Optic NoIx Int Int Int Int` will complain for more local kind error.

phadej:
Ok. I think the discussion is converged to the consensus, that `*` is good enough.

PR #56, opened 2019-01-13
Implement Mapping using roam

phadej:


arybczak:
Can you explain the motivation behind these?

phadej:
`map'` -> `roam`:

- Main motivation is `inspection-testing`; with definition using `map'`; the computation nrule doesn't hold. (I I read the core right, using `Context` made something stricter)
- It's consistent with `Traversing` wander

Adding `linear` is similar motivation, including that conversion from van Laarhoven representation doesn't involve `Context` (in most cases).

arybczak:
Ok. Can you make the utility function for one particular definition of `linear` that's repeated 4 times?

arybczak:
Also INLINE is missing on iroam definitions.

phadej:
> Ok. Can you make the utility function for one particular definition of linear that's repeated 4 times?

Is it ok to make default implementation?

PR #59, opened 2019-01-14
Change abstract interface

### LINKED ISSUE #41: Make an Optic interface a little more abstract (hide indexed implementation details)

phadej:
Resolves #41

I temporarily moved `ixcontramap` to `Traversing` (from `Profunctor` where it belongs).

### DIFF
--- a/optics-core/src/Optics/Internal/Optic/TypeLevel.hs
+++ b/optics-core/src/Optics/Internal/Optic/TypeLevel.hs
@@ -47,7 +47,6 @@ instance AppendProof xs ys zs => AppendProof (x ': xs) ys (x ': zs) where
   appendProof i = case appendProof @xs @ys @zs i of
     Refl -> Refl
 
-{-
 
 class CurryCompose xs where
   composeN :: (i -> j) -> Curry xs i -> Curry xs j
@@ -60,6 +59,7 @@ instance CurryCompose xs => CurryCompose (x ': xs) where
   composeN ij f x = composeN @xs ij (f x)
   {-# INLINE composeN #-}
 
+{-
 
 class ConstN xs where
   constN :: y -> Curry xs y
--- a/optics-core/src/Optics/Internal/Optic/TypeLevel.hs
+++ b/optics-core/src/Optics/Internal/Optic/TypeLevel.hs
@@ -1,0 +1,73 @@
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE InstanceSigs #-}
+module Optics.Internal.Optic.TypeLevel where
+
+import Data.Type.Equality
+import Data.Kind (Type)
+import Data.Proxy (Proxy (..))
+
+-- | Curry a type-level list.
+--
+-- In pseudo (dependent-)Haskell:
+--
+-- @
+-- 'Curry' xs y = 'foldr' (->) y xs
+-- @
+type family Curry (xs :: [*]) (y :: *) :: * where
+  Curry '[]       y = y
+  Curry (x ': xs) y = x -> Curry xs y
+
+-- |
+--
+-- 'AppendProof' is a very simple class which provides a witness
+--
+-- @
+-- foldr f (foldr f init xs) ys = foldr f init (ys ++ xs)
+--    where f = (->)
+-- @
+--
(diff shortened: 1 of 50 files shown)


optics-core/src/Optics/Internal/Optic/TypeLevel.hs
```diff
@@ -0,0 +1,73 @@
+{-# LANGUAGE TypeApplications #-}
```
arybczak: Can we keep 2 spaces for indentation everywhere?

optics-sop/src/Generics/SOP/Optics.hs
```diff
@@ -21,49 +21,49 @@ import Generics.SOP
 import Optics hiding (to)
 
 -- | Iso between a generic type and its representation.
-rep :: (Generic a, Generic b) => Iso i a b (Rep a) (Rep b)
+rep :: (Generic a, Generic b) => Iso a b (Rep a) (Rep b)
 rep = iso from to
 
--- | Iso induced by the 'SOP' newtype.
-sop :: Iso i (SOP f xss) (SOP g yss) (NS (NP f) xss) (NS (NP g) yss)
+-- | Isonduced by the 'SOP' newtype.
```
arybczak: It seems you search-and-replaced a bit too much here ;) Same below.

phadej: Thanks, fixed

arybczak:
Latest commit changes functions that expect a regular optic to functions that take any optic, ie. both regular and indexed ones, silently ignoring the indices in the latter case. Do we really want it? So far we agreed for explicit conversion only.

phadej:
We agreed with @kosmikus and @adamgundry, that this is ok. Hopefully there weren't any miscommunication.

It's easy to change though with `CheckNonIndexed` class.

adamgundry:
Yes, we have a design choice here, but our most recent thinking is that we should permit using an indexed optic as a regular one, rather than mandating an explicit conversion. This fits with the perspective that indices are somewhat orthogonal to the optic kind: an indexed setter is "just" a setter that can provide indices, but they don't necessarily have to be used. Thus we allow silent downgrading much as we allow a lens to be silently used as a setter, even though it is capable of more.

There doesn't seem to be much harm to the extra flexibility; it allows things like `toListOf ifolded` where `toListOf folded` would do, which might be undesirable in some cases, but in others it is nice not to require separate names for the indexed and non-indexed versions.

@arybczak are there any major downsides of this approach that we are missing?

arybczak:
Alright, I just wanted to make sure we're on the same page.

The only thing I can think about is the fact that indexed versions with unused indices are slower than regular ones. But perhaps we can mitigate that by coming with something similar (yet simpler) to `Conjoined` in the future.

phadej:
Rebased on current `master`.

PR #65, opened 2019-01-14
No backticks oleg

phadej:
#52 with exported `OpticKind` from internal module.


arybczak:
At this point I vote for reverting the OpticKind change and use empty data types instead.

I'd like the type of `_1 % _1 to` be `(Field1 s t u v, Field1 u v a b) => Optic A_Lens o o s t a b`, not `(Field1 s t u v, Field1 u v a b) => Optic 'Optics.Internal.OpticKind.A_Lens o o s t a b`, the module prefix adds too much noise IMO. And we don't really get anything practically useful for using OpticKind anyway, it's not like someone will end up with `Optic Int` by accident.

PR #68, opened 2019-01-15
Add proof-of-concept generic-optics

phadej:


arybczak:
Ha, label version looks great. We also avoid the issue described in https://github.com/kcsongor/generic-lens/issues/16#issuecomment-344404893 by Optic being a newtype instead of type synonym.

adamgundry:
Yes, that's the idea. Indeed the genesis of this library was the problem that `lens`-style optics don't play nicely with `IsLabel`, so it's nice to see this finally coming together. The `HasField` built-in class should eventually provide an alternative to using generics to get the getter/setter, but it's nice that we can essentially do without it.

phadej:
I briefly discussed with Csongor, and he'll try to make a way to use `generic-lens` to generate profunctor optics. Then we get sane way to do all what `generic-lens` can, e.g. traversals etc.

But for now, I'm going to merge this as is so it don't bit rot.

arybczak:
We don't need `generic-lens` to provide profunctor lenses, VL lenses and traversals are already trivial to wrap with `lensVL` and `traversalVL`, whereas `prismVL` can be written cheaply and without much trouble too ('ll add it later to `generic-optics`).

phadej:
They *are*, but it's unneeded conversion back and fortth as `generic-lens` internals are profunctor based.

Issue #78, opened 2019-01-16
Expose CheckIndices for documentation purposes and cleaner signatures


arybczak:
Currently we have this:

```
λ> :i iover
iover ::
  (optics-core-0.1:Optics.Internal.Indexed.CheckIndices i is,
   Is k A_Setter) =>
  Optic k is s t a b -> (i -> a -> b) -> s -> t
        -- Defined in ‘optics-core-0.1:Optics.Internal.IxSetter’
```

Also haddock doesn't let you go to definition of CheckIndices. We can expose CheckIndices, but make it have a hidden superclass so that it's not possible to add instances.

PR #79, opened 2019-01-16
Expose CheckIndices for haddock and cleaner signatures

arybczak:
Solves #78.

I skipped the internal superclass because it actually makes things worse.

```haskell
λ> :t itraverseOf
itraverseOf
  :: (Optics.Internal.Indexed.CheckIndicesInternal i,
      Is k A_Traversal, Applicative f) =>
     Optic k '[i] s t a b -> (i -> a -> f b) -> s -> f t
```

I don't think it's a big deal though as this class is only for helpful error messages.

### DIFF
--- a/optics-core/optics-core.cabal
+++ b/optics-core/optics-core.cabal
@@ -84,6 +84,7 @@ library
 
   default-extensions:
     BangPatterns
+    DataKinds
     DefaultSignatures
     DeriveFunctor
     FlexibleContexts
@@ -97,5 +98,3 @@ library
     TupleSections
     TypeFamilies
     TypeOperators
-
-  default-extensions: DataKinds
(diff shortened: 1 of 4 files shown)


optics-core/optics-core.cabal
```diff
@@ -84,6 +84,7 @@ library
 
   default-extensions:
     BangPatterns
+    DataKinds
```
phadej: :+1:

optics-core/src/Optics/Internal/Indexed.hs
```diff
@@ -61,7 +60,7 @@ instance
   ) => CheckIndices i '[i1, i2, i3, i4, i5]
 
 instance
-  ( TypeError ('Text "Use icompose* variants to flatten indices")
+  ( TypeError ('Text "Use icomposeN to flatten indices")
```
phadej: we don't export `icomposeN`, do we.

I think we could, but then we have to export the `CurryCompose` family (or at least mention that it has instances for all type level lists).

arybczak: Alright, I reverted this change to error message for now, we can do that later.

PR #80, opened 2019-01-16
Port Data.Data.Optics from lens

arybczak:
Sans `upon` and its variations because of https://github.com/ekmett/lens/issues/235.

TH generation uses `Control.Lens.Plated`. Luckily it uses instances with default implementation, i.e. `uniplate`, which `Data.Data.Optics` provides.

Btw @phadej, your reorganization was a bit premature, because TH generation uses stuff that you've put into `optics` package, so I'm just going to stick everything into `optics` until we have working TH generation and then we can move things around with more confidence.

### DIFF
--- a/optics/optics.cabal
+++ b/optics/optics.cabal
@@ -26,15 +26,16 @@ library
   hs-source-dirs:     src
   ghc-options:        -Wall
   build-depends:
-      array           >=0.5.1.1 && <0.6
-    , base            >=4.9     && <5
-    , containers      >=0.5.7.1 && <0.7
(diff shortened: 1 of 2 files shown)


phadej:
Please make an issue with checkbox-list of *what* should be moved *where* (and *why* it cannot be put there right now).

I'd rather put TH dependencies into `optics-th` and see how much stuff ends up there, and then think how to push things out, if they are generally usable.

arybczak:
Actually, scratch that, the code in this module is super opaque and has weird hidden issues such as https://github.com/ekmett/lens/issues/742.

PR #83, opened 2019-01-19
TH optics generation

arybczak:
I think some minor documentation changes are needed and ThTest.hs needs to be made a proper testsuite, but other than that it's complete.

I adjusted lens code to generate affine folds and traversals when appropriate.


arybczak:
Haddock on Travis hangs with GHC 8.2.2 because of https://github.com/haskell/haddock/issues/679 - not sure how to properly require a newer one in this case.

PR #84, opened 2019-01-19
Make (i)traverse(Of)_ more efficient, add some documentation

arybczak:


arybczak:
I added https://github.com/well-typed/optics/commit/9d92d9ed0aafd7672b4b475a76f98b76be63b336 later, so ignoring the value is baked in `runTraversed`.

PR #89, opened 2019-01-20
Expose internal modules of optics-core

arybczak:


phadej:
I don't like this change. This should been discussed.

PR #94, opened 2019-01-20
Implement equivalent of Conjoined from lens for efficient optics degradation

arybczak:
Woohoo!

Now we get degradation of indexed optics without loss of efficiency in comparison to regular versions. And the implementation is surprisingly simple.

### DIFF
--- a/optics-core/src/Optics/Internal/Profunctor.hs
+++ b/optics-core/src/Optics/Internal/Profunctor.hs
@@ -256,66 +256,78 @@ class (Choice p, Strong p) => Traversing p where
   wander
     :: (forall f. Applicative f => (a -> f b) -> s -> f t)
     -> p i a b -> p i s t
-
   iwander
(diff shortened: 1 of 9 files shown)


optics-core/src/Optics/Internal/Profunctor.hs
```diff
@@ -256,66 +256,77 @@ class (Choice p, Strong p) => Traversing p where
   wander
     :: (forall f. Applicative f => (a -> f b) -> s -> f t)
     -> p i a b -> p i s t
-
   iwander
     :: (forall f. Applicative f => (i -> a -> f b) -> s -> f t)
     -> p j a b -> p (i -> j) s t
 
+  conjoined
```
phadej: Note: this method can be in `Profunctor`.

arybczak: I'll add a comment. We can move it along with ixcontramap, as only then it'll actually be useful there.

Issue #30, opened 2019-01-07
And Ixed and At classes


phadej:
If we depend on `containers` these would provide tools to work with maps.

Unlike `lens`, `Ixed` could provide **affine traversal**.

PR #97, opened 2019-01-21
Drop support for GHC 8.0.2 and add 'both'

arybczak:
PR to test waters, but it looks like it simplifies things quite a lot.


adamgundry:
I'm in two minds about this... on the one hand I can see that it makes things a fair bit simpler. On the other, I've been trying out `optics` on a project that currently still needs to support 8.0.2.

arybczak:
Was the tryout successful? ;)

adamgundry:
Broadly successful in that I got the library to compile with `optics` without needing much more than mechanical changes. I did come across a few missing pieces on which my code relies, which are currently stubbed out and raised as issues (#30, #103).

arybczak:
Alright, let's keep the compatibility for now. it's not a big deal.

Issue #57, opened 2019-01-13
Rename `ViewableOptic` to simply `View` / consider its validity


phadej:

kosmikus:
Is there any true benefit to having the `View(ableOptic)` generalisation over just having the `view1`, `view01`, and `viewN` functions to use which are more precise?

arybczak:
API compatibility with `lens` and `^.` that "just works".

kosmikus:
I'm not yet convinced this is an advantage in this case.

arybczak:
FWIW I'm pretty sure that if we don't have `.^` (which is most likely (one of) the most used operators in lens) the stream of complaints will be endless.

kosmikus:
Oh, I think we should have `^.`, but I think it should be equivalent to `flip view1`.

arybczak:
I have an idea.

What about making view for Fold and Traversal be toListOf?

That's consistent with other implementations of view - Lens/Getter give you `r`, Prism/AffineTraversal/AffineFold give you `Maybe r`, it's only logical that Traversal/Fold give you `[r]`. 

I think one of the first mistakes people make with `lens` is that they try to view a traversal or fold and expect a list, but instead they get that confusing error message about `no Monoid instance for` (or completely unexpected result if the return type happens be a Monoid). That change will make it work as expected.

arybczak:
Another possibility is to scrap ViewableOptic class altogether and:
- rename `view1` to `view`
- rename `view01` to `preview` and remove original `preview` that acts on folds. We can later add `firstOf` that does the same thing (but more efficiently, according to lens documentation).
- rename `viewN` to `foldOf`

arybczak:
I'm pretty convinced that the above is the move to go for, except we should also keep ViewableOptic class with current polymorphic `view` (after renaming it to something like `gview`) as it's needed for things like `passthrough` and `use`.

arybczak:
> Another possibility is to scrap ViewableOptic class altogether and:
> 
>     * rename `view1` to `view`
> 
>     * rename `view01` to `preview` and remove original `preview` that acts on folds. We can later add `firstOf` that does the same thing (but more efficiently, according to lens documentation).
> 
>     * rename `viewN` to `foldOf`

> I'm pretty convinced that the above is the move to go for, except we should also keep ViewableOptic class with current polymorphic `view` (after renaming it to something like `gview`) as it's needed for things like `passthrough` and `use`.

Both things were done.

PR #109, opened 2019-01-26
Make view return a list of results for (indexed) traversals and folds

arybczak:
Alright, nobody protested after I posted https://github.com/well-typed/optics/issues/57#issuecomment-456911279 so here's an implementation for testing the waters.

I'm also exporting `Optics.Operators `as part of `Optic.Core`. It currently has 6 operators:
- `&` - reexported from base
- `<&>` - reexported from base
- `^.` - with this change it somewhat becomes "one operator to view them all", i.e. it just works for majority of optics and gives you the result you'd expect (especially with traversals and folds). As a bonus you can run it for indexed variants and get indices.
- `#` - primarily for reverse application of `Iso` and `Prism`
- `%~` and `.~` - along with `&` used for nice update/set syntax: `T{_age=5} & age %~ (+1)`

I think they are core and should be exported by default.

I also moved `^..` and `^?` to `Optics.Operators.Compat` as I think `^.` is sufficient. `^..` and `^?` actually give you an illusion of type safety - they return `[a]` and `Maybe a` respectively, but they work for any fold. That means that if you refactor the code and e.g. your prism becomes a fold, return type of `^?` will not change and you might lose results if you miss it. This won't happen with `^.` because its return type will change from `Maybe a` to `[a]` and GHC will complain.

Demonstration:
```haskell
λ> Left "hi" ^. _Left
Just "hi"
λ> (1,2) ^. _1
1
λ> [Left ('c', "hi"), Right 42, Left ('a', "there")] ^. traversed % _Left % _2
["hi","there"]
λ> [Left ('c', "hi"), Right 42, Left ('a', "there")] ^. itraversed % _Left % _2
[(0,"hi"),(2,"there")]
```

### DIFF
--- a/optics-core/src/Optics/Core.hs
+++ b/optics-core/src/Optics/Core.hs
@@ -21,6 +21,7 @@ import Optics.Indexed         as O
 import Optics.Iso             as O
 import Optics.Lens            as O
 import Optics.LensyReview     as O
+import Optics.Operators       as O
 import Optics.Passthrough     as O
 import Optics.Prism           as O
 import Optics.PrismaticGetter as O
--- a/optics-core/src/Optics/Operators.hs
+++ b/optics-core/src/Optics/Operators.hs
@@ -38,7 +38,7 @@ infixl 1 <&>
 #endif
 
 -- | Flipped infix version of 'view'.
-(^.) :: ViewableOptic k a => s -> Optic' k is s a -> ViewResult k a
+(^.) :: ViewableOptic k is a => s -> Optic' k is s a -> ViewResult k is a
 (^.) = flip view
 {-# INLINE (^.) #-}
 
--- a/optics-core/src/Optics/Operators.hs
+++ b/optics-core/src/Optics/Operators.hs
@@ -58,7 +58,7 @@ infixl 8 ^..
 
 infixl 8 ^?
 
--- | Flipped infix version of 'review'.
+-- | Infix version of 'review'.
 (#) :: Is k A_Review => Optic' k is t b -> b -> t
 (#) = review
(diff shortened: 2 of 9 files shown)


arybczak:
No one has any opinions about this?

phadej:
If `view` on `Fold` would return `[a]`, so should `viewN` (and then we'll need `foldOf`),

I also like an idea of making

```haskell
(^.) = flip view1
(^?) = flip view01
(^..) = flip viewN -- the [a] returning one
```

---

I'm :-1: on re-exporting operators.

arybczak:
> If `view` on `Fold` would return `[a]`, so should `viewN` (and then we'll need `foldOf`),

Hmm, yes, that makes sense.

> I also like an idea of making
> 
> ```haskell
> (^.) = flip view1
> (^?) = flip view01
> (^..) = flip viewN -- the [a] returning one
> ```

That works fine, but you lose the ability to use indexed variants (lens has different set of operators for indexed versions, namely `^@.`, `^@?`, `^@..`, which are redundant with `(^.) == flip view`).

> I'm -1 on re-exporting operators.

Why? There are 4 optics-specific ones that simplify common usage a lot.

`Lens.Operators` has 122, which is of course overwhelming, but 4 is IMO completely fine.

phadej:
I see. With this PR `view` (and `(^.)` if it's `= flip view`) indeed has super powers. But then it's weird, shouldn't `%~` be the same then (i.e. make `%@~` redundant).

arybczak:
We could add equivalent `SettableOptic` class.

arybczak:
Closing in favor of #113.

PR #111, opened 2019-02-02
Split optics-extra, port bytestring/text/vector stuff from lens

### LINKED ISSUE #47: Add Each Text and Each ByteString instances
Also `packed` and `unpacked` are useful.

I think they can be in `optics-core`.

### LINKED ISSUE #64: Proper credit to Edward
but remove module headers

### LINKED ISSUE #81: Another package named e.g. `optics-extra`
I think there needs to be another package between `optics-core` and `optics-th` that contains stuff that TH uses, but also pulls in a couple more dependencies for completeness of instances.

E.g. TH needs `Data.Data.Optics` (#80), but it:
- has nothing to do with TH
- requires `unordered-containers`

TH generation also needs
- `Control.Lens.At` - Not yet ported, I think can be worked around.
- `Control.Lens.Wrapped` - This is used for generation of `Wrapped` instances for newtypes. We can skip it for now. The code is mostly trivial, but it has tons of dependencies.
- `Data.Tuple.Optics` - We can provide local versions of these if we really want.

Currently `Optics.Each` and indexed optics are missing instances for `unordered-containers` (and lesser extent `text`, `bytestring`, `vector`). I think it would be good to fit these into `optics-extra` (along with `Data.Data.Optics`) and add missing instances.

Or somewhere.

arybczak:
Fixes #64, fixes #81, fixes #47.

optics-core now only depends on `base`, `array` and `containers`. I left `*WithIndex` in optics-core since documentation uses `itraversed` etc. in examples and the majority of instances can be there anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,10 +68,10 @@ install:
   - "sed -i.bak 's/^jobs:/-- jobs:/' ${HOME}/.cabal/config"
   - rm -fv cabal.project cabal.project.local
   - grep -Ev -- '^\s*--' ${HOME}/.cabal/config | grep -Ev '^\s*$'
-  - "printf 'packages: \"optics\" \"optics-core\" \"optics-sop\" \"optics-th\" \"generic-optics\" \"metametapost\"\\n' > cabal.project"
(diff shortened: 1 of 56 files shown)


arybczak:
Documentation needs a little polish I think, but that can be done later.

PR #112, opened 2019-02-03
Add VL affine traversal and Port Optics.At from lens

### LINKED ISSUE #30: And Ixed and At classes

arybczak:
Fixes #30.

The first commit borrows some ideas from #98 to get affine traversals in VL form which are usually easier to define than using `atraversal` and can be more efficient.

### DIFF
--- a/optics-core/src/Optics/AffineTraversal.hs
+++ b/optics-core/src/Optics/AffineTraversal.hs
@@ -26,6 +26,11 @@ type AffineTraversal s t a b = Optic An_AffineTraversal NoIx s t a b
 type AffineTraversal' s a = Optic' An_AffineTraversal NoIx s a
 
 -- | Type synonym for a type-modifying van Laarhoven affine traversal.
+--
+-- Note: this isn't exactly van Laarhoven representation as there is no
+-- @PointedFunctor@ class, but you can interpret the first argument as a
+-- dictionary of 'Pointed' class that supplies the @point@ function.
+--
 type AffineTraversalVL s t a b =
   forall f. Functor f => (forall r. r -> f r) -> (a -> f b) -> s -> f t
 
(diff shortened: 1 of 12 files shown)


arybczak:
I also threw in `Optics.Cons` as I tried it and it compiled without much intervention.

optics-core/src/Optics/AffineTraversal.hs
```diff
@@ -21,6 +25,13 @@ type AffineTraversal s t a b = Optic An_AffineTraversal NoIx s t a b
 -- | Type synonym for a type-preserving affine traversal.
 type AffineTraversal' s a = Optic' An_AffineTraversal NoIx s a
 
+-- | Type synonym for a type-modifying van Laarhoven affine traversal.
+type AffineTraversalVL s t a b =
+  forall f. Functor f => (forall r. r -> f r) -> (a -> f b) -> s -> f t
```
phadej: this is not the van Laarhoven representation. Correct one is `forall f. PointedFunctor f => ...`. This is naming issue, so some other name should be used for this type alias

arybczak:
I actually had Empty already ported locally, just didn't push it.

optics-core/src/Optics/AffineTraversal.hs
arybczak: True. I figured it's close enough though. It's a similar problem to `foldVL` function.

I'm not sure what the name should be though. Any suggestions?

arybczak: I created #114 for tracking this. It can stay like that for the moment, I'd like to have access to affine traversal specific stuff from this PR in #113 to figure out `zoom`, so I'm going to merge this tomorrow.

Issue #103, opened 2019-01-23
Add zoom and use functions


adamgundry:
I don't have a very good feel for these yet, but they came up in some code I was porting from `lens`.

arybczak:
#113 defined `use` in a way such that it works as both `use` and `preuse` from the lens library.

In theory if we make gview return multiple results with folds and traversals, we can also make it work like `ipreuse` and `iuse` from lens (well, actually better, because it won't return monoidal summary of indices when used with indexed fold, which as the lens library highlights doesn't make sense).

Alternative is to have all of these variations explicit (which I guess is consistent with view/preview/toListOf family). @adamgundry what do you think?

I'm not sure how `zoom` fits into the picture, but if it won't need `gview`, we can merge parts of ViewableOptic into `PermeableOptic` and get rid of `gview` entirely. But maybe it's alright to keep it... Ahh, so many questions.

PR #113, opened 2019-02-03
view -> foldOf, view1 -> view, view01 -> preview, generalized ViewableOptic

arybczak:
Improved version of #109.

Now:
- `view` / `^.` is for getters,
- `preview` / `^?` is for affine folds,
- `toListOf` / `^..` is for folds,
- `viewN` is called `foldOf`

and ViewableOptic has `gview` that does crazy things to allow concise implementation of `passthrough` (and `use` etc. in the future).


arybczak:
I backed off from the idea of returning multiple results with `gview`. I think as of now these changes are fairly uncontroversial.

arybczak:
I've added `use`, `uses`, `listening` and `listenings`. `pre` versions are not required.

Eh, but perhaps we should just have `use` for getters, `preuse` for affine folds and nothing for folds?

There is also `zoom`, but I didn't figure out its story yet.

If we don't need `gview` for `zoom`, perhaps we can get just get rid of it? Oleg, what about `passthrough`, how would that work without `gview`?

arybczak:
I'm going to merge this (while adding a note that the whole View module is experimental for now), the discussion about what `use` should ultimately be can continue in #103.

Issue #115, opened 2019-02-05
Strict functions for setters, space leaks tracking


arybczak:
I skimmed over open issues for lens and the following mention space leaks problems:

1. [Add strict functions](https://github.com/ekmett/lens/issues/574)
2. [Add a strict version of %~](https://github.com/ekmett/lens/issues/664)
3. [Space leak in partsOf](https://github.com/ekmett/lens/issues/219)
4. [Add a class for certain indexed efficiencies](https://github.com/ekmett/lens/issues/790)

(1) and (2) are easily solvable by adding appropriate variants.

(3) doesn't seem to happen with latest lens. We don't have `partsOf` yet, but I wrote a base implementation a while ago:
```haskell
partsOf :: forall k s t a. Is k A_Traversal => Optic k NoIx s t a a -> Lens s t [a] [a]
partsOf o = lensVL $ \f s ->
  let as = toListOf (Optic (rphantom . getOptic (toTraversal o)) :: Fold s a) s
      update a = get >>= \case
        []       ->            pure a
        a' : as' -> put as' >> pure a'
  in evalState (traverseOf o update s) <$> f as
```
which is more efficient than the lens one (both space and time wise). The one from lens  essentially does the same thing if you look past the abstractions, but I guess they confuse the optimizer too much. The lens implementation is of course a bit more polymorphic as it also accepts folds and turns them into getters, but this variation could arguably be just named differently (if we want to provide it at all).

As for (4), I think we don't have such issues because we never pair indices and values. @treeowl do you have any examples that leak space with indexed optics from lens so that we can check that we're issues free?

arybczak:
`at` for Map can introduce unwanted laziness: https://github.com/ekmett/lens/issues/492

arybczak:
Added `at'` in f5c678166546c333e850579558fe9f9b7133c6c7.

arybczak:
Added another `at'` (for `IntMap`) in bec6a7aebf4b012af8abad757f4643a12dc88052.

arybczak:
Closing as we now have `at'`, `set'`, `over'` and strict versions of operators.

PR #116, opened 2019-02-05
Add strict versions of (i)over and (i)set

arybczak:
Addresses (1) and (2) of #115.

Should we have infix versions of `over'` and `set'` (`%!~` and `!~`)?

### DIFF
--- a/optics-core/src/Optics/Internal/Profunctor.hs
+++ b/optics-core/src/Optics/Internal/Profunctor.hs
@@ -8,6 +8,7 @@ module Optics.Internal.Profunctor where
 import Data.Coerce (Coercible, coerce)
 import Data.Functor.Const
 import Data.Functor.Identity
+import Data.Semigroup
 
 import Optics.Internal.Utils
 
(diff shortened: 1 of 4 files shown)


optics-core/src/Optics/Internal/Profunctor.hs
```diff
@@ -472,20 +498,36 @@ instance Traversing IxFunArrow where
...
 
 instance Mapping FunArrow where
-  roam f (FunArrow k) = FunArrow (f k)
+  roam  f (FunArrow k) = FunArrow $ f k
+  iroam f (FunArrow k) = FunArrow $ f (const k)
   {-# INLINE roam #-}
+  {-# INLINE iroam #-}
 
-  iroam f (FunArrow k) = FunArrow (f (const k))
+instance Mapping (IxStar ((,) Unit')) where
```
phadej: This is ugly. I'd guess, that `StrictFunArrow` is not much more code.

arybczak: Unfortunately not, you need to write instances for all profunctor classes (times two, because there is also StrictIxFunArrow).

We get these for free with `(Ix)Star` because we have instances for `Applicative f => (Ix)Star f` below `Mapping`.

arybczak: Alternative is Star StrictIdentity, but that type has unlawful Applicative instance (see https://github.com/ekmett/lens/issues/664#issuecomment-234434845). I don't know if that matters though.

arybczak: I'm not even sure if it's possible to do this with FunArrow.

`StrictFunArrow` is equivalent to `Star StrictIdentity`, so in pseudo-Haskell:

```haskell
newtype StrictFunArrow i a b = StrictFunArrow { runStrictFunArrow :: a -> !b }
```

when you have `StrictFunArrow k`, you need to force the result of application of `k`. But how can you do it when writing e.g. the definition of `roam`? `k` is just passed to `f` there, you can't influence what happens with the result of `k` applied to the value we're modifying.

phadej: So the point is to `seq` only the fields focused by optic. `StrictIdentity` would force whole structure, won't it? Am I overthinking?

This is highly non-trivial, so proper `[Note]` should be in place.

phadej: Yes, I now see that we need a token, forcing which will force proper parts.

But it's something one need to think about when seen first time, it looks too complicated. Don't make adam or andres or possible contributors to wonder what happens, but write a note. This is tricky, and should be clarified.

arybczak: > So the point is to `seq` only the fields focused by optic. `StrictIdentity` would force whole structure, won't it? Am I overthinking?

It would force everything that's on the rhs of `<*>` in the definition of traversals. Exactly which parts would be forced depends on the way the traversal is written, which is quite bad.

> This is highly non-trivial, so proper `[Note]` should be in place.

Yes, that indeed needs to be explained.

arybczak: I replaced `Unit'` with proper `Identity'` type so that its inner workings are less obscure and added explanation. It should be fairly clear now what's going on.

PR #119, opened 2019-02-07
Move appropriate parts of At, Cons, Each and Empty to optics-core

arybczak:
So that we can have all the core classes in optics-core as well as better `doctest` examples.

It increases compilation time by ~0.7 seconds for me, so it's pretty insignificant.


phadej:
IIRC, @kosmikus opinion (which I not disagree with), is that convenience type-classes don't belong to `optics-core`.

With https://github.com/sol/doctest/pull/217 you can run `doctest` cross package suite: i.e. symbol defined in `optics-core` can use the whole `Optics` for doctest, if it is desirable.

arybczak:
> IIRC, @kosmikus opinion (which I not disagree with), is that convenience type-classes don't belong to `optics-core`.

I think @kosmikus just wanted to keep the list of dependencies small, these don't add anything in this regard.

> 
> With [sol/doctest#217](https://github.com/sol/doctest/pull/217) you can run `doctest` cross package suite: i.e. symbol defined in `optics-core` can use the whole `Optics` for doctest, if it is desirable.

It's one thing to be able to run `doctest` cross-package and another having a user read documentation and think "why does this example use functions that are not defined here?"

Advantages to including these in optics-core:
- documentation examples use known functions, so you can just get optics-core and play with thngs
- it allows other packages to add instances for these without having to depend on optics-extra, which has more overhead

There are no disadvantages except a very slightly increased compilation time and a couple more orphan instances, but we already have them with `*WithIndex` classes, so it's a non-issue.

phadej:
My understanding, is that we'd recommend to use `optics` package (i.e. `cabal new-repl optics` if you want to play); `optics-core` is an option for those who care about dependency footprints.

Thus, adding type-classes to `optics-core` is a double-edged sword.
- It's nice they are there, so I can depend only on `optics-core` in say something like `insert-ordered-containers` to provide `At`.
- OTOH, that we'll lead to increasing scope leak of `optics-core`, as essentially everything without dependencies blogs there.
    - E.g. we can separate `optics-indexed` out of `optics-core`; especially as internals are exposed
    - but if the policy is "everything without dependencies goes into `optics-core`" then we don't need `optics-indexed` for anything.

arybczak:
As for the optics-indexed, I silently agreed with what Adam suggested in https://github.com/well-typed/optics/issues/81#issuecomment-456314082, i.e. separation of optics-indexed and optics-extra doesn't really buy us much, but adds maintenance overhead (and similar documentation issues as highlighted above, i.e. we use itraversed/ifolded/imapped in optic-core examples).

The primary reason for changes in this PR is to make it a possibility for libraries to depend on optics-core, not the other way around.

There is of course a lot of uncertainty whether this project will gain traction and talking about wide adoption (because only then most library authors will even consider depending on it) is ambitious, but if At/Cons/Each/Empty are not in optics-core and having them requires non-trivial dependencies, then providing instances for them most likely won't be considered at all.

PR #121, opened 2019-02-09
Port parts of generic-lens needed for overloaded labels

arybczak:
This is somewhat experimental, but I wanted to check how well we do at non-trivial things, both feature and performance wise. It seems we do a good job.

```haskell
λ> view #_2 ("hi", "there")
"there"
λ> view #_3 ("hi", "there")

<interactive>:3:6: error:
    • The type ([Char], [Char]) does not contain a field at position 3
    • In the first argument of ‘view’, namely ‘#_3’
      In the expression: view #_3 ("hi", "there")
      In an equation for ‘it’: it = view #_3 ("hi", "there")
λ> preview #_Left (Left 'a')
Just 'a'
λ> preview #_Letf (Left 'a')

<interactive>:18:1: error:
    • The type Either
                 Char b does not contain a constructor named "Letf"
        it :: forall b a.
              (Optics.Internal.Generic.Sum.Constructors.GAsConstructor
                 "Letf"
                 (M1
                    C
                    ('MetaCons "Right" 'PrefixI 'False)
                    (S1
                       ('MetaSel
                          'Nothing 'NoSourceUnpackedness 'NoSourceStrictness 'DecidedLazy)
                       (Rec0 b)))
                 (M1
                    C
                    ('MetaCons "Right" 'PrefixI 'False)
                    (S1
                       ('MetaSel
                          'Nothing 'NoSourceUnpackedness 'NoSourceStrictness 'DecidedLazy)
                       (Rec0 b)))
                 a
                 a,
               (TypeError ...)) =>
              Maybe a
λ> import GHC.Generics
λ> data T = T1 { tint :: Int } | T2 { tint :: Int, tstring :: String } deriving (Show, Generic)
λ> let t = T2 50 "foo"
λ> view #tint t
50
λ> view #tin t

<interactive>:19:6: error:
    • The type T does not contain a field named 'tin'.
    • In the first argument of ‘view’, namely ‘#tin’
      In the expression: view #tin t
      In an equation for ‘it’: it = view #tin t
λ> preview #_T2 t
Just (50,"foo")
λ> preview #_T1 t
Nothing
```

Code generation is very good, we get equivalents of hand-written lenses and prisms, the exceptions being:
- GHC 8.0 works fine if it comes to lenses for one constructor data types, but it doesn't optimize generics away for sum types
- With GHC 8.4 and lower prisms such as `#_T2` don't fully optimize away the HList (used internally) in the `construct` function (GHC 8.6 does).

For example:

```haskell

data X = X1 Int | X2 String String
  deriving Generic

rrr :: Prism' X (String, String)
rrr = #_X2
```

GHC 8.6.3 generates the following constructor:

```haskell
-- RHS size: {terms: 15, types: 32, coercions: 0, joins: 0/0}
rrr2
rrr2
  = \ ds1 ->
      case ds1 of {
        Left x -> x;
        Right y ->
          X2 (case y of { (a1, b1) -> a1 }) (case y of { (a1, b1) -> b1 })
      }
```

whereas GHC 8.4.4 and earlier:

```haskell
-- RHS size: {terms: 27, types: 138, coercions: 11, joins: 0/1}
rrr2
rrr2
  = \ ds1 ->
      case ds1 of {
        Left x -> x;
        Right y ->
          let {
            ds
            ds = case y of { (a1, b1) -> (a1, :> @~ <Co:7> b1 $WNil) } } in
          X2
            (case ds of { (a1, b1) -> a1 })
            (case ds of { (a1, b1) ->
             case b1 of { :> @ a2 @ as cobox x1 ds2 -> x1 `cast` <Co:4> }
             })
      }
```


arybczak:
I added tests.

arybczak:
Closing in favor of #132.

PR #127, opened 2019-02-14
Make each supply indexed traversals

arybczak:


adamgundry:
Should `Each` now have additional functional dependencies `s -> i, t -> i`? Otherwise we can end up with ambiguity when `each` is composed.

arybczak:
Yes, I updated fundeps in https://github.com/well-typed/optics/commit/c0ffb6f3a0bcd6707185f663ce64e39aaca3791e.

PR #100, opened 2019-01-22
WIP: conjoined without internals

phadej:

PR #138, opened 2019-02-24
Add indexed affine fold/traversal, add conjoined to IxOptic

arybczak:
Finishes #98 and #100.


arybczak:
:(

Alright, I'll just merge then.

PR #142, opened 2019-03-11
Rework diagram generation; Use hackage version of topograph

phadej:


metametapost/src/MetaMetaPost.hs
```diff
@@ -267,6 +262,12 @@ data St s = St
     , stVars  :: PerTy Int
     }
 
+instance (a ~ b, a ~ ([Stmt s] -> [Stmt s])) => LabelOptic "stmts" A_Lens (St s) (St s) a a where
```
arybczak: `b` is only present in the context, same in the instance below.

PR #145, opened 2019-03-13
WIP: Docs improvements

adamgundry:
This structures the documentation for each kind of optic around formation/introduction/elimination rules, adds some (brief) comments for each one and fixes lots of Haddock links.

I'm not sure whether it is worth breaking out computation rules as a separate section, and the docs are not entirely consistent yet. Perhaps the haddocks for the "core" introduction/elimination forms should simply state the rules they satisfy. Alternatively we could distinguish between "core" and "derived" introduction/elimination forms.

How to arrange the docs between the different `optics-*` modules needs some thought, in particular because `reexported-modules` don't seem to show up in Haddock. For example, we should perhaps move the diagrams to `optics-core` (as well?). 

This still needs some higher-level introduction and motivation, and an explanation of how indexed optics work.

### DIFF
--- a/optics-core/src/Optics/Fold.hs
+++ b/optics-core/src/Optics/Fold.hs
@@ -428,8 +428,7 @@ sumOf :: (Is k A_Fold, Num a) => Optic' k is s a -> s -> a
 sumOf o = foldlOf' o (+) 0
 {-# INLINE sumOf #-}
 
--- | The sum of a collection of actions, generalizing 'concatOf'.
--- TODO: we don't yet have 'concatOf'!
+-- | The sum of a collection of actions.
 --
 -- >>> asumOf both ("hello","world")
 -- "helloworld"
@@ -444,7 +443,7 @@ asumOf :: (Is k A_Fold, Alternative f) => Optic' k is s (f a) -> s -> f a
 asumOf o = foldrOf o (<|>) empty
 {-# INLINE asumOf #-}
 
--- | The sum of a collection of actions, generalizing 'concatOf'.
+-- | The sum of a collection of actions.
 --
 -- >>> msumOf both ("hello","world")
 -- "helloworld"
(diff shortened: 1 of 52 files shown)


arybczak:
Thanks a lot for this!

BTW, should we agree on the column width for documentation comments for consistency? I think you use C-q to align them in emacs and you have default value of `fill-column` (70), though I personally think that's a bit too little and have it set to 80. The result is that whenever you or I try to edit comments written by one another, the whole paragraph will either be shrunk to 70 or expanded to 80 characters depending on who wrote it.

adamgundry:
> BTW, should we agree on the column width for documentation comments for consistency?

You guess correctly, and I'm quite happy to change `fill-column` to 80.

optics-core/src/Optics/Fold.hs
```diff
@@ -395,6 +429,7 @@ sumOf o = foldlOf' o (+) 0
 {-# INLINE sumOf #-}
 
 -- | The sum of a collection of actions, generalizing 'concatOf'.
+-- TODO: we don't yet have 'concatOf'!
```
arybczak: I didn't add `concatOf` because `concatOf` is a restricted version of `foldOf` (though the comment slipped through).

optics/src/Optics.hs
```diff
@@ -205,10 +208,34 @@ module Optics
...
 
-  -- * At
+  -- * Optics utilities
+
+  -- ** At
+
+  -- | An 'AffineTraversal' to traverse a key in a map or an element of a
+  -- sequence:
+  --
+  -- >>> headOf (ix 1) ['a','b','c']
```
arybczak: Using `preview` would be more clear here :)

optics-core/src/Optics/Fold.hs
adamgundry: Fixed by removing the reference.

optics/src/Optics.hs
adamgundry: Thanks, good point. That's what I get for trying to think of examples late at night.

arybczak:
Do you think we can merge the PR and start another one for fine-tuning? It'll allow us to avoid any conflicts as the PR touches a lot of things.

adamgundry:
Yes, good plan. I'm still occasionally polishing things and will try to gather some feedback, but in the meantime I think we are better off merging this.

PR #147, opened 2019-03-13
Cleanups

phadej:


arybczak:
Generated file `optics.png` contains indexed affine fold and traversal, it shouldn't ;)

arybczak:
Fixed in 003beec.

PR #152, opened 2019-04-19
Remove Equality optic

### LINKED ISSUE #151: So we need `Equality`
If we remove that, then `Optic` will always be morally coercible in `s t a b`.
I didn't ever used `Equality`, `simple` could be an `Iso` too.

```haskell
newtype Optic (k :: *) (is :: [*]) s t a b =
  Optic { getOptic :: forall p j. Profunctor p => Optic_ k p j (Curry is j) s t a b }
```

which will make all optics `CoercibleOptics`. And `Ixed`, if/when `conjoined'` is moved to our `Profunctor`

arybczak:
Fixes #151.


cabal.project
```diff
@@ -6,6 +6,10 @@ packages:
   optics-th/*.cabal
   generic-optics/*.cabal
 
+-- Code generation for optic kinds
+packages:
+  codegen/*.cabal
```
phadej: I specifically have `cabal.codegen.project`, so one don't need to support wide GHC window for utilities

adamgundry:
This seems fine to me, at least, I've never found myself using `Equality`.

As a consequence of this `Is k An_Iso` is morally equivalent to `k ~ An_Iso`. Should we use this fact to simplify the type signatures of `Iso` elimination forms? I don't have a strong opinion either way... using the `Is` constraints feels more consistent, but the specialised types might give fractionally better type inference.

arybczak:
I removed
* codegen from cabal.project
* `Is k An_Iso` constraints from `Optics.Iso` module

Thanks for suggestions, I'll merge the PR later.

PR #153, opened 2019-04-30
Add IxLens and IxGetter

phadej:
Note: `conjoined` and `icomposeN` could be written for all optic kinds now, as there aren't `Equality`.


arybczak:
> `conjoined` and `icomposeN` could be written for all optic kinds

Yes, that is a plus. On the other hand, these type variables...

I really am not sure if it's worth adding this.

I think two questions are relevant:
1) How useful (if at all) are indexed lenses? Can you give an example?
2) Can these coindices be (somewhat) easily made to offer the functionality of what you showed in your blog post, e.g. making prisms that give you info about failure in such a way that's compatible (and composable) with other optic kinds?

phadej:
My main motivation is the symmetry.

I'll answer *How useful (if at all) are indexed lenses?* with another question: How useful is `LensyReview`? If we remove `Costrong` and `Cochoice`, and implement `re` with functional dependency class `Getter <-> Review` and `Iso <-> Iso`, then we can also get rid of second pair of indices.

My gut feeling is that `IxLens` is more useful than `PrismaticGetter`.

phadej:
`selfIndex :: IxGetter i i i` is useful.

arybczak:
I'm not a fan of `LensyReview` and `PrismaticGetter` either, but you can largely ignore their existence, whereas these two type variables in `Optic__` are always there :(

phadej:
yet, `Optic__` is an implementation detail ;)

arybczak:
@adamgundry what do you think? I am somewhat torn on this.

adamgundry:
I don't have a very strong grasp on this, so I'm not sure I should have the casting vote. :wink:

> `selfIndex :: IxGetter i i i` is useful.

If so, we should export it from `Optics.IxGetter`... and I think if we are to have `IxGetter` and `IxSetter` then having `IxLens` is better than not having it, in the interests of consistency.

As far as I can see the only impact on the public API is the presence of `IxGetter`/`IxSetter`, right? I'm not so worried about two more type variables in internal types, provided that the public API remains relatively simple (and in particular, that those who aren't interested in indexed optics can safely ignore them).

I'm very interested in the answer to @arybczak's question 2 above, i.e. how easy is this to extend to "coindexed prisms"? I guess that does involve further change to the public API, but might also be a more compelling justification?

Regarding `LensyReview` and `PrismaticGetter`, I was meaning to start a separate discussion on what we should do with them. I think they are something of a misnomer (and misleadingly documented at present, probably by me): a `LensyReview` is really a "backwards `Lens`" just as a `Review` is a "backwards `Getter`". If we're going to keep them, we should be more explicit and have canonical introduction/elimination rules for them. But I'm not totally convinced that we shouldn't just drop them and limit `re` to `Getter`/`Review`/`Iso`.

phadej:
The *coindexed prisms* is a beast I barely know. If it's something that let's return something on non-match; then I'm not sure this PR will help with it.

We use `forall R. I -> R ~ I` (yoneda/cps) and thus `(I, J) ~ forall R. (I -> J -> R)` isos to make indexes (their composition) nicer. But it feels that this blocks from using coindices in a sum-way.

I'd imagine that coindexed prism implementation would change

```haskell
newtype ForgetM r a b = ForgetM (a -> Maybe r)
-- to
newtype ForgetM r js a b = ForgetM (a -> Either (NS I (() ': js)) r)
```

I.e. coindexes would allow to "fail" with a message. But that `NS` stuff doesn't seem to fit `optics` at all. (It's ok on a "theory"-side. though)

arybczak:
After giving it some thought I'd say it's fine to add this, but I'd like to have type variable that correspond to coindices be something different than `l`. How about `ci`/`cj`? Then it's much easier to see and remember these are (co)indices.

Also, please add `selfIndex` to `IxGetter`.

phadej:
The index is unchanged in almost every use case, thus `ci/cj` feels not optimal. I settled with `l` as it didn't clash with anything, but it could be **i**ndex, and coinde**x**? :)

arybczak:
If coindex is not changed, it can be `ci`/`ci`. **i**ndex/**c**o**i**ndex is better, because `x` can refer to both index and coindex.

arybczak:
I also can see that they are sometimes denoted as `l` and sometimes as `j`. This is really confusing. We should use for them the same i/j as for indices, but prefixed with `c`, hence `ci`/`cj` is the best.

arybczak:
@phadej can you rename coindex variables appropriately (i.e. `ci` etc. everywhere) so that we can merge this?

phadej:
@arybczak I don't have time to work on `optics` atm.

arybczak:
I renamed the variables and made `conjoined` and `icomposeN` standalone functions.

adamgundry:
Thanks to both of you for taking care of this. @arybczak are you still planning to add `selfIndex`? Otherwise, to the extent that I understand this, I think it is now good to merge?

arybczak:
We can add `selfIndex` separately, indeed I think it's alright to merge.

arybczak:
Let's not merge yet, I need to check something.

arybczak:
Ok, so coindex variables are needed only for type tetris in Re.hs for defining total instances.

But note the type signature of `re` is `Optic k NoIx s t a b -> Optic (ReversedOptic k) NoIx b a t s`: it only works for optics without indices. Also note that the only way to drop indices is through `noIx`, which reinterprets the optic, so that index-specific class functions are not used anymore.

Therefore the definitions of index-specific class functions (`ixcontramap`, `ilinear`) in instances for `Re` are never actually used, so in principle they can be" implemented" using `error` and it'll work fine.

Issue #20, opened 2018-11-24
Alternatives for % as optic composition (and its fixity)


arybczak:
The main problem is that it conflicts with Data.Ratio. Did you consider any alternatives?

kosmikus:
I think we're quite happy to discuss other options, if you have any suggestions.

arybczak:
Also, see a reddit comment of Edward I dug up: https://www.reddit.com/r/haskell/comments/13l9vj/getting_started_with_lenses/c75w6ap/

> Until lens 3.4 this was a consistent application of the above principle, applying a function `(%) :: a -> (a -> b) -> b`.
>
> So knowing one led you to the other. In 3.4 we switched to `(&)` after complaints, because it was causing issues with traditional Haskell golf code that involved `Data.Ratio` on lambdabot.

phadej:
I was accidentally looking into "what to use as a composition combinator" (for one different, opaque encoding of optics), and I was thinking about `%` (single character) and `.>` too.

I guess we can have `.>` as the default. Only practical use can show how annoying (or not) is to not have single character combinator.

adamgundry:
Sounds like `.>` is the way to go.

kosmikus:
I'm ok with `.>` as well.

phadej:
thinking a bit more: `o .> o'` (`lens`) is `icompose (\i j -> j) (o % o')` (current `optics`)

Also we could/should have indexedgetter, indexedaffinetraversal etc; so I think that indexed vs. non indexed is another dimension of optics lattice; not really a part of it. (every non-indexed thing is indexed thing with a trivial index)

Let's not change the operator just yet.

arybczak:
> thinking a bit more: `o .> o'` (`lens`) is `icompose (\i j -> j) (o % o')` (current `optics`)

Yeah, but that is just a consequence of how indexed optics behave in lens, i.e. only leftmost index is preserved by default.

> Also we could/should have indexedgetter, indexedaffinetraversal etc; so I think that indexed vs. non indexed is another dimension of optics lattice; not really a part of it. (every non-indexed thing is indexed thing with a trivial index)

I'm not sure about this one. Anything above traversal/fold gives you zero or one result, what is the point of indexing it?

phadej:
> I'm not sure about this one. Anything above traversal/fold in the hierarchy gives you zero or one result, what is the point of indexing it?

[...]

PR #154, opened 2019-04-30
WIP: filtered is AffineFold

phadej:


arybczak:
Looks good.

Let's replace `filtered` from `Optics.Fold` with this one.

BTW, they are combinators simply because you can't write `ifiltered` in the other form as you don't have access to the index, so I made `filtered` have the same form for consistency.

As a bonus, I'm pretty sure that if `unsafeFiltered` is a combinator, you can precisely write when it's a valid traversal.

arybczak:
Oh wait, I didn't actually notice that you moved away from the combinator style. Right, so I'd like to preserve it for reasons mentioned above.

adamgundry:
I don't have a strong opinion on combinator vs compositional style, as I haven't thought about the problem, but if we end up keeping the combinator style we should document the divergence from `lens`.

phadej:
For the record: I tried to make machinery to make `ifiltered`-like combinators be definable "inline", i.e. I tried to make following possible to write

```haskell
myfold = ifolded %? ifiltered (\i a -> ...)
```

1. it requires making `%` **infixl** (which I think we *should do* that)
2. adding new type: `OpticIxTransformer k is js s t a b` (note the `js`)
    - handwavy intuition: `Optic` *is* `OpticIxTransformer` which may only add indices, ~`Monoid`: `Optic is ~ forall js. OpticIxTransformer js (is ++ js)`. The restrictiveness of current `Optic` **is** good thing, but there are (e.g. `ifiltered`) where it's unfortunate.
3. but unfortunately the current machinery (= implementation of indices) resisted this frankly
    - I will try slightly different implementation as an "research" experiment, but I don't suggest changing what we have now.

Even unsuccessful, this small exercise made me want `%` and `%%` to be `infixl`; that's how I read the optics composition already, and optic-transformer-not-combinator `ifiltered` would work. It cannot be pre-composed. We'd want following operators (below) and `%%?` **is not** associative, which suggest "correct" associativity for `%%`.

```haskell
(%%)  :: Optic k is s t u v -> Optic              k    js u v a b -> Optic k (is ++ js) s t a b
(%%?) :: Optic k is s t u v -> OpticIxTransformer k is js u v a b -> Optic k        js  s t a b
```
```

arybczak:
:+1: for making `%` infixl. I think we should just do it as it's more natural and nobody objected to the idea in #20 . Would you mind creating a PR for that?

arybczak:
Since `ifiltered` can't be made an affine fold, let's close this.

PR #155, opened 2019-04-30
Move matching and isn't to AffineTraversal

### LINKED ISSUE #148: Move matching and isn't to AffineTraversal?
Is there any reason not to define `matching` and `isn't` on affine traversals rather than prisms? That would allow them to be used on the composition of a lens and a prism, which is nice.

I came across this while working on #145, and am tempted to include it in that to save an unnecessary conflict.

adamgundry:
Fixes #148. This makes the elimination/computation rules for `AffineTraversal` cleaner.

### DIFF
--- a/optics-core/src/Optics/AffineFold.hs
+++ b/optics-core/src/Optics/AffineFold.hs
@@ -25,7 +25,10 @@ module Optics.AffineFold
   -- 'preview' ('afolding' f) ≡ f
   -- @
 
-  -- * Semigroup structure
+  -- * Additional elimination forms
+  , isn't
+
+    -- * Semigroup structure
   , afailing
 
   -- * Subtyping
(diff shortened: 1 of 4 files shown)


arybczak:
Ah, I see that `matching` requires affine traversal because of `Either t a`. But `isn't` doesn't, maybe it's worth making `isn't` work on affine folds?

adamgundry:
That would make sense, although I'm wondering if it is even worth having both `hasn't` for folds and `isn't` for affine folds/traversals? I suppose the naming does suggest the difference between the possibility of many elements and the possibility of one element, but given that they are functionally equivalent, should we just drop `isn't`?

arybczak:
I think it's fine to keep `isn't` as we have `preview` and `headOf` (although this is a bit different case, but still).

adamgundry:
Ok, I've moved `isn't` to `AffineFold`. Assuming Travis is happy I think this is now good to merge?

arybczak:
Looks good :+1:

PR #168, opened 2019-05-29
Increase bounds on th-abstraction

Avi-D-coder:
The [breaking change in `0.3`](https://github.com/glguy/th-abstraction/blob/master/ChangeLog.md) appears to be irrelevant as `datatypeVars` is not used.

> Breaking change: the datatypeVars field of DatatypeInfo is now of type [TyVarBndr] instead of [Type], as it now refers to all of the bound type variables in the data type. The old datatypeVars field has been renamed to datatypeInstTypes to better reflect its purpose.


arybczak:
Looks good, thanks!

PR #171, opened 2019-05-30
Make filtered inline, add unsafeFiltered and indices

arybczak:
I added `unsafeFIltered` and `indices`. I also made `filtered` inline so that it's compatible with `lens`.

`unsafeFiltered` and `filtered` can be defined as affine traversal and fold, but I didn't do that because they have at most one target, so it doesn't make much sense to me

### DIFF
--- a/optics-core/src/Optics/Fold.hs
+++ b/optics-core/src/Optics/Fold.hs
@@ -42,6 +42,7 @@ module Optics.Fold
   , folding
   , foldring
   , unfolded
+  , filtered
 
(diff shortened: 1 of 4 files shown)


phadej:
if `filtered` is inline, it's ought to be `AffineFold`

arybczak:
@phadej I thought it doesn't make much sense as affine fold has at most one target, but hmm. Then you also want to make `unsafeFiltered` affine traversal for consistency and the description of why it's unsafe gets a bit weird.

PR #173, opened 2019-05-30
Add %& post-composition operator.

phadej:
Allows to put filtered, ifiltered and other such not-quite-optics
into the optic composition chain.

left-associativity is essential for this to work,
but it works nicely!

### DIFF
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -26,6 +26,7 @@ module Optics.Internal.Optic
   , castOptic
   , (%)
   , (%%)
+  , (%&)
   , IsProxy(..)
   -- * Labels
   , LabelOptic(..)
@@ -36,6 +37,7 @@ module Optics.Internal.Optic
   , module Optics.Internal.Optic.TypeLevel
   ) where
 
+import Data.Function ((&))
 import Data.Proxy (Proxy (..))
 import Data.Type.Equality
 import GHC.OverloadedLabels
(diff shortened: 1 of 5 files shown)


arybczak:
Thanks, it's interesting that it works out. We should add example to documentation of `ifiltered`/`indices` so that people don't miss it.

adamgundry:
I'm impressed that you managed to find an operator that `lens` isn't using... it seems `optics` may be growing its own little family of composition operators... :wink: But indeed, let's do this and mention this on the Haddocks for cases where it is useful.

optics-core/src/Optics/Internal/Optic.hs
```diff
@@ -152,6 +154,22 @@ Optic o %% Optic o' = Optic oo
       (o . o')
 {-# INLINE (%%) #-}
 
+-- | Flipped function application, specialised to optics and binding tightly.
+--
+-- Useful for post-composing optics transformations:
+--
+-- >>> toListOf (folded %& filtered odd) [1..10]
```
phadej: argh, this is now broken as filtered is an optic...

phadej:
Examples added (and fixed) to at least some candidates.

optics-core/src/Optics/IxFold.hs
```diff
@@ -192,6 +196,8 @@ ifiltered
 ifiltered p o = mkIxFold $ \f ->
   itraverseOf_ o (\i a -> if p i a then f i a else pure ())
 {-# INLINE ifiltered #-}
+-- Note: technically one could filter lens into atraversal, atraversal into
+-- itself filtering atraversal might be useful in fact.
```
arybczak: I don't understand this note, can you clarify?

phadej: that's the same idea that `filtered` is an affine fold.

arybczak: Ok, but it looks to me like there is some punctuation missing there, I can't parse that sentence :c

PR #174, opened 2019-05-30
ReversedOptic is an injective type family

adamgundry:
Without this we have

```hs
set @A_Lens . re
  :: (ReversibleOptic k, ReversedOptic k ~ A_Lens) =>
     Optic k NoIx b a t s -> b -> s -> t
```

but adding it gives

```hs
set @A_Lens . re :: Optic A_LensyReview NoIx b a t s -> b -> s -> t
```


arybczak:
Good observation :+1:

PR #175, opened 2019-05-30
Document LensyReview and PrismaticGetter properly

### LINKED ISSUE #169: LensyReview and PrismaticGetter
From https://github.com/well-typed/optics/pull/153#issuecomment-489461017:

> Regarding `LensyReview` and `PrismaticGetter`, I was meaning to start a separate discussion on what we should do with them. I think they are something of a misnomer (and misleadingly documented at present, probably by me): a `LensyReview` is really a "backwards `Lens`" just as a `Review` is a "backwards `Getter`". If we're going to keep them, we should be more explicit and have canonical introduction/elimination rules for them. But I'm not totally convinced that we shouldn't just drop them and limit `re` to `Getter`/`Review`/`Iso`.

Thinking about this a bit further, it would be odd from a subtyping perspective if `re` didn't work on a `Lens` but did on an `Iso`. So I suppose our options are to properly document and add canonical forms for `LensyReview`/`PrismaticGetter`, or to have `re` of a `Lens` return a `Review`. I'm slightly inclined to the first, but what do others think?

If we do keep them, I'd suggest more explicit names, perhaps `ReversedLens` and `ReversedPrism`? (Just as `Review` could be `ReversedGetter`, not that I'm suggesting renaming it.)

adamgundry:
Fixes #169, modulo a decision about renaming. I'm inclined to go for `ReLens` and `RePrism` but that will have to wait for another evening.

### DIFF
--- a/optics-core/src/Optics/LensyReview.hs
+++ b/optics-core/src/Optics/LensyReview.hs
@@ -1,18 +1,53 @@
 -- |
 -- Module: Optics.LensyReview
--- Description: A 'Optics.Review.Review' produced by 'Optics.Re.re' on a 'Optics.Lens.Lens'.
+-- Description: A backwards 'Optics.Lens.Lens'.
 --
--- A 'LensyReview' is a 'Optics.Review.Review' produced by calling
--- 'Optics.Re.re' on a 'Optics.Lens.Lens'.  It is essentially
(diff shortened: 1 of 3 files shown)


arybczak:
LGTM, thanks :+1:

PR #181, opened 2019-06-04
Regenerate .travis.yml; add GHC-8.8.1 job

phadej:


phadej:
I got travis to the state where it fails on our code with GHC-8.8.1. That's good enough for me.

PR #182, opened 2019-06-05
Support for turning read-write optics into their read-only counterparts

arybczak:
It's the equivalent of `getting` function from lens.

I didn't find `getting` to be a good name, but I'm not sure if `readOnly` is good either - suggestions welcome.

### DIFF
--- a/optics-core/src/Optics/ReadOnly.hs
+++ b/optics-core/src/Optics/ReadOnly.hs
@@ -8,32 +8,47 @@ import Optics.Internal.Profunctor
 
 -- | Class for read-write optics that have their read-only counterparts.
 class ToReadOnly k where
-  type ReadOnlyOptic k
   -- | Turn read-write optic into its read-only counterpart.
-  getting :: Optic k is s t a b -> Optic' (ReadOnlyOptic k) is s a
+  --
+  -- This is useful when you have an @optic :: Optic k is s t a b@ of read-write
+  -- kind @k@ such that @s@, @t@, @a@, @b@ are rigid, there is no evidence that
+  -- @s ~ t@ and @a ~ b@ and you want to pass @optic@ to one of the functions
+  -- that accept read-only optic kinds.
+  --
+  -- Example:
+  --
+  -- @
+  -- λ> let fstIntToChar = _1 :: Lens (Int, r) (Char, r) Int Char
+  -- λ> :t view fstIntToChar
+  --
+  -- <interactive>:1:6: error:
+  --     • Couldn't match type ‘Char’ with ‘Int’
+  --       Expected type: Optic' A_Lens NoIx (Int, r) Int
+  --         Actual type: Lens (Int, r) (Char, r) Int Char
+  --     • In the first argument of ‘view’, namely ‘fstIntToChar’
+  --       In the expression: view fstIntToChar
+  -- λ> :t view (getting fstIntToChar)
+  -- view (getting fstIntToChar) :: (Int, r) -> Int
+  -- @
+  getting :: Optic k is s t a b -> Optic' (Join A_Getter k) is s a
 
 instance ToReadOnly An_Iso where
-  type ReadOnlyOptic An_Iso = A_Getter
   getting o = Optic (getting__ o)
   {-# INLINE getting #-}
 
 instance ToReadOnly A_Lens where
-  type ReadOnlyOptic A_Lens = A_Getter
   getting o = Optic (getting__ o)
   {-# INLINE getting #-}
 
 instance ToReadOnly A_Prism where
-  type ReadOnlyOptic A_Prism = An_AffineFold
   getting o = Optic (getting__ o)
   {-# INLINE getting #-}
 
 instance ToReadOnly An_AffineTraversal where
-  type ReadOnlyOptic An_AffineTraversal = An_AffineFold
   getting o = Optic (getting__ o)
   {-# INLINE getting #-}
 
 instance ToReadOnly A_Traversal where
-  type ReadOnlyOptic A_Traversal = A_Fold
   getting o = Optic (getting__ o)
   {-# INLINE getting #-}
 
--- a/optics-core/src/Optics/ReadOnly.hs
+++ b/optics-core/src/Optics/ReadOnly.hs
@@ -7,8 +7,9 @@ import Optics.Internal.Optic
 import Optics.Internal.Profunctor
 
 -- | Class for read-write optics that have their read-only counterparts.
-class ToReadOnly k where
-  -- | Turn read-write optic into its read-only counterpart.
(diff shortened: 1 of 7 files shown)


phadej:
IMO if there aren't clearly better name, than sticking to `lens`-name is less confusing.

optics-core/src/Optics/ReadOnly.hs
```diff
@@ -0,0 +1,46 @@
...
+
+import Optics.Internal.Bi
+import Optics.Internal.Optic
+import Optics.Internal.Profunctor
+
+-- | Class for read-write optics that have their read-only counterparts.
+class ReadOnly k where
+  type ReadOnlyOptic k
+  -- | Turn read-write optic into its read-only counterpart.
+  readOnly :: Optic k is s t a b -> Optic' (ReadOnlyOptic k) is s a
```
phadej: isn't this composition with  `to id :: Getter a a`?

phadej: And if it is, then `readOnly :: Getter a b -> Getter a b` etc. should be valid thing to do.

arybczak: It's not, the type signature of `\o -> o . to id` is too restricted.

arybczak:
> IMO if there aren't clearly better name, then sticking to lens-name is less confusing.

Ok, so leave the class name as-is, but make the function `getting`?

optics-core/src/Optics/ReadOnly.hs
phadej: Then it should be explained in the comment, as I find it hard to believe.

phadej:
I agree, `Getting` would be terrible class name, so it can stay `ReadOnly` or maybe `ToReadOnly`.

optics-core/src/Optics/ReadOnly.hs
arybczak: You don't need to believe it, you can see for yourself.

```haskell
λ> :t \o -> to id % o
\o -> to id % o
  :: (Is l (Join A_Getter l), Is A_Getter (Join A_Getter l)) =>
     Optic l js v v a b -> Optic (Join A_Getter l) js v v a b
λ> :t \o -> o % to id
\o -> o % to id
  :: (Is k (Join k A_Getter), Is A_Getter (Join k A_Getter)) =>
     Optic k is s t b b -> Optic (Join k A_Getter) is s t b b
```

arybczak:
I renamed `ReadOnly` to `ToReadOnly` and `readOnly` to `getting` :+1:

optics-core/src/Optics/ReadOnly.hs
phadej: Now I don't understand:

```haskell
Prelude Control.Lens> toListOf (getting traverse) "foo"
"foo"
```

How that works, as AFAICS, `b` is not solved, and is ambiguous?

---

Some use example where indices of `Optic k is s t a b` cannot vary would be illustrative.

---

Anyway, my comment was related to the fact that in `lens` you can do:

```haskell
Prelude Control.Lens> view (getting (to id)) 'x'
'x'
```

So we should have `ReadOnly A_Getter` etc. instances for already readonly optics too (which would be (some) coerce).

arybczak: > How that works, as AFAICS, b is not solved, and is ambiguous?
I'm not sure what you mean.

```haskell
λ> :t getting
getting
  :: ToReadOnly k =>
     Optic k is s t a b -> Optic' (ReadOnlyOptic k) is s a
λ> :t getting traversed
getting traversed :: Traversable t => Optic' A_Fold NoIx (t a) a
λ> :t getting itraversed
getting itraversed
  :: TraversableWithIndex i f => Optic' A_Fold (WithIx i) (f a) a
λ> :t getting (itraversed % itraversed)
getting (itraversed % itraversed)
  :: (TraversableWithIndex i1 f1, TraversableWithIndex i2 f2) =>
     Optic' A_Fold '[i1, i2] (f1 (f2 a)) a
```

`getting` coerces `Optic k is s t a b` to `Optic k is s s a a` (using `Bicontravariant`) so `t` and `b` are irrelevant, they are ignored.

arybczak: > So we should have ReadOnly A_Getter etc. instances for already readonly optics too (which would be (some) coerce).

Right, we can add instances for read-only optics that don't do anything to the optic.

adamgundry:
Some examples of when `getting` is necessary would be nice. I guess in principle I can imagine situations where there is an ambiguous variable otherwise, but I'm not sure how they arise in practice.

optics-core/src/Optics/ReadOnly.hs
adamgundry: Will it at least be the case that `ReadOnlyOptic k` is always `Join A_Getter k`? If so, do we need the extra type family?

arybczak:
I added example usage and removed `ReadOnlyOptic`.

optics-core/src/Optics/ReadOnly.hs
```diff
@@ -8,32 +8,47 @@ import Optics.Internal.Profunctor
...
+  -- λ> :t view fstIntToChar
+  --
+  -- <interactive>:1:6: error:
+  --     • Couldn't match type ‘Char’ with ‘Int’
+  --       Expected type: Optic' A_Lens NoIx (Int, r) Int
+  --         Actual type: Lens (Int, r) (Char, r) Int Char
+  --     • In the first argument of ‘view’, namely ‘fstIntToChar’
+  --       In the expression: view fstIntToChar
+  -- λ> :t view (getting fstIntToChar)
+  -- view (getting fstIntToChar) :: (Int, r) -> Int
```
adamgundry: Thanks, this is a really helpful example.

Though it does make me wonder... the problem here could be solved by giving `view` a more general type, using `Optic` rather than `Optic'`. Is there a reason we shouldn't do that? It introduces two more type variables, of course, so perhaps they might sometimes be ambiguous?

phadej: @adamgundry they will be ambiguous

arybczak: Actually I tried to make all of the read-only optic specific functions take and return `s t a b` instead of `s s a a`, though eventually I got ambiguous type variable errors in some auxiliary module (I think it was `Optics.Cons.Core`)

arybczak:
Do we want `ToReadOnly` instances for read-only optic kinds?

phadej:
@arybczak I cannot think any harm in having the instances, so let's add them.

arybczak:
Alright, I added missing instances. It makes the class definition a bit uglier, but oh well.

PR #180, opened 2019-06-01
Add optics-compat for compatibility layer between van Laarhoven optics

arybczak:
Fixes #165.

PR #184, opened 2019-06-05
Split optics-th into two packages

### LINKED ISSUE #141: optics-th should not use TH itself

### LINKED ISSUE #159: Split optics-th into two packages?
See https://github.com/ekmett/lens/issues/844 for background. We should probably aim to release with GHC 8.8 support.

adamgundry:
Fixes #141 and fixes #159.

This adds `optics-template-haskell` as a package containing optics for the `template-haskell` library (I'm happy to take name suggestions). The `optics-th` package now contains a small subset of `Language.Haskell.TH.Optics`, which is no longer exported by it or the main `optics` package.

Having tried this I think it is worth doing: the `optics-th` package now requires no CPP and compiles with the latest `template-haskell`, whereas `optics-template-haskell` doesn't (yet) because it needs yet more CPP.

This does lead to some duplication of code in both `optics-th` and `optics-template-haskell`. An alternative would be to have a dependency one way or the other, but neither seems entirely satisfactory.


phadej:
`optics-template-haskell` →`template-haskell-optics` which will follow the (until now) unwritten guide:
- `$pkg-optics` for optics for types in $pkg (e.g. `template-haskell-optics`, or `these-optics`, or `lens-optics`)
- `optics-$modifier` for some additional functionality

Note: how `optics-sop` is two purpose package as well. Compared to `generics-sop-lens` and `lens-sop`.

adamgundry:
I like that naming scheme, although would it imply that `optics-vl`(#180) should be renamed to `lens-optics`?

phadej:
@adamgundry it doesn't even depend on `lens`, does it ;)

arybczak:
Do we still include `template-haskell-optics` as a dependency of `optics` or make it optional?

Making it optional will solve the problem of having to bump version of `optics` after each GHC release, so it would be more convenient.

adamgundry:
I originally left it out (and hence this PR changes `optics.cabal` to no longer reexport `Language.Haskell.TH.Optics`). I'm inclined to think that is the right trade-off, in the interests of keeping `optics` as stable as possible over multiple GHC releases.

Good plan re versioning `template-haskell-optics` in line with `template-haskell`.

adamgundry:
Just for the record: we decided not to bump the version number of `template-haskell-optics` as the current version works across several `template-haskell` versions, not just the latest.

PR #185, opened 2019-06-08
Compare performance with lens in benchmarks

arybczak:
Looks like in the vast majority of cases we're either on par or slightly above `lens` in performance.


arybczak:
There is nothing weird in the code changes, so I'm just gonna merge it.

PR #189, opened 2019-06-10
Fix spelling in example

Avi-D-coder:
Is it mentioned anywhere that `FlexibleInstances`, `MultiParamTypeClasses`, `UndecidableInstances` are needed? Should I enable them by default, when making `optics`?


arybczak:
For `makeFieldLabels` you need DataKinds, FlexibleInstances, MultiParamTypeClasses, TypeFamilies and UndecidableInstances in modules where you generate them.

For `makeFields` you need FlexibleInstances  and MultiParamTypeClasses.

For `makeLenses` you shouldn't need anything.

PR #192, opened 2019-06-15
Licensing/authorship

### LINKED ISSUE #167: Authors/maintainers in .cabal files
I don't think the maintainer email address `optics@well-typed.com` exists yet (unless @kosmikus has set it up). We should either set it up or change it. @arybczak do you have an opinion on what this should be?

For the authors lists, I'm not sure about including the original authors of code we have ported, at least without asking first. We should certainly credit them appropriately in the description/README, but listing as an author could be taken to imply more support for `optics` than we can necessarily assume...

Are there any other `.cabal` metadata fields that need attention?

adamgundry:
This corrects attribution in the `.cabal` files and updates the licenses, including with the licenses of packages from which we have derived code. Fixes #167.

### DIFF
--- a/generic-optics/LICENSE
+++ b/generic-optics/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2017-2018, Well-Typed LLP
+Copyright (c) 2017-2019, Well-Typed LLP
 
 All rights reserved.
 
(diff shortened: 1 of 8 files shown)


arybczak:
LGTM :+1:

PR #201, opened 2019-06-18
Fix doctest examples

arybczak:
It would be good to also make CI run `make doctest`, but I'm not sure how to do it.


arybczak:
Yes, let's do that :+1:

PR #203, opened 2019-06-18
Add function equivalents of operators in Optics.Operators.State

### LINKED ISSUE #162: Add function equivalents of operators in Optics.Operators.State
`assign`, `modifying` etc.

arybczak:
Fixes #162.

### DIFF
--- a/optics-extra/src/Optics/State.hs
+++ b/optics-extra/src/Optics/State.hs
@@ -1,0 +1,80 @@
+-- |
+-- Module: Optics.State
+-- Description: 'Setter' utilities for working with 'MonadState'.
+--
+module Optics.State
+  ( modifying
+  , modifying'
+  , assign
+  , assign'
+  ) where
+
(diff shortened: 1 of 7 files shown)


optics-extra/src/Optics/State.hs
```diff
@@ -0,0 +1,66 @@
...
+modifying
+  :: (Is k A_Setter, MonadState s m)
+  => Optic k is s s a b
+  -> (a -> b)
+  -> m ()
+modifying o = modify . over o
+{-# INLINE modifying #-}
+
+-- | Version of 'modifying' that is strict in both optic application and state
+-- modification.
```
adamgundry: It would be really nice to write down the strictness properties and/or give examples of the difference. (This applies to e.g. `set'` as well.)

arybczak: I added examples.

PR #205, opened 2019-06-20
Move LabelOptic to Optics.Label and work on haddocks

adamgundry:


optics-core/src/Optics/Label.hs
```diff
@@ -0,0 +1,201 @@
...
+--
+-- Perhaps Peter is going on vacation and needs to leave his pets at home:
+--
+-- >>> peter & set #pets []
+-- Human {humanName = "Peter", humanAge = 13, humanPets = []}
+--
+--
+-- == Structure of 'LabelOptic' instances
+--
+-- /Note:/ You might wonder why instances above are written in form
```
arybczak: `/Note:/` is not needed here as it's the beginning of a new section.

arybczak:
Looks good (modulo one comment), thanks!

optics-core/src/Optics/Label.hs
adamgundry: Good point, removed. I'm not totally satisfied with the docs here but I don't plan to do any more on them right now.

PR #210, opened 2019-06-27
Per-optic-kind diagrams

### LINKED ISSUE #158: Diagrams of parts of subtype hierarchy
We could consider making additional versions of the [nice diagram](https://github.com/well-typed/optics/blob/master/optics/optics.png), one for each optic kind, that highlights that kind and shows its subtypes and supertypes. These could then be included in the subtyping section of the per-kind pages.

adamgundry:
Fixes #158. This adds a diagram for each (non-indexed) optic kind, highlighting that kind and those above and below it in the partial order. @phadej's metametapost code was nicely easy to extend. :grinning:

Which colours should we paint this family of bikesheds? I'm not especially attached to the current colour scheme.

Where should these appear in the docs? At the moment I have them in the "subtyping" section under the tag type.


optics-core/src/Optics/Review.hs
```diff
@@ -25,6 +25,7 @@ module Optics.Review
 
   -- * Subtyping
   , A_Review
+  -- | <<Review.png Review in the optics hierarchy>>
   , toReview
```
adamgundry: Looks like `toReview` crept through when we culled the other similar functions?

phadej:
Nice idea to simply gray out stuff, i.e. play only with colors! Looks great, both the code is simple and image results are nice.

adamgundry:
> Nice idea to simply gray out stuff, i.e. play only with colors! Looks great, both the code is simple and image results are nice.

Yes, that was @arybczak's idea on #158, I just hacked it into your code. :wink:

arybczak:
Looks great, nice that it didn't take a lot of code to extend image generation.

As for colors, I prefer what I used in #158 - black only and lighter grey seems more readable to me.

I didn't even notice there were two shares of blue before I looked at the code :P


```diff
diff --git a/metametapost/src/MMP/Optics/Hierarchy.hs b/metametapost/src/MMP/Optics/Hierarchy.hs
index 91d2d61..b6e9b31 100644
--- a/metametapost/src/MMP/Optics/Hierarchy.hs
+++ b/metametapost/src/MMP/Optics/Hierarchy.hs
@@ -78,8 +78,8 @@ hierarchyFocus focus = hierarchyColoured (text_colour . fromEnum)
 
     text_colour k
       | k == f           = red
-      | G.path graph f k = light_blue
-      | G.path graph k f = dark_blue
+      | G.path graph f k = black
+      | G.path graph k f = black
       | otherwise        = grey
 
     arrow_colour j k
@@ -87,9 +87,7 @@ hierarchyFocus focus = hierarchyColoured (text_colour . fromEnum)
       | G.path graph j f, G.path graph k f = black
       | otherwise                          = grey
 
-red, light_blue, dark_blue, grey, black :: Expr s 'Color
-red        = RGB (L 1)   (L 0)   (L 0)
-light_blue = RGB (L 0)   (L 0)   (L 0.75)
-dark_blue  = RGB (L 0)   (L 0)   (L 0.5)
-grey       = RGB (L 0.5) (L 0.5) (L 0.5)
-black      = RGB (L 0)   (L 0)   (L 0)
+red, grey, black :: Expr s 'Color
+red        = RGB (L 0.75) (L 0)    (L 0)
+grey       = RGB (L 0.75) (L 0.75) (L 0.75)
+black      = RGB (L 0)    (L 0)    (L 0)
```

Lens

optics-core/src/Optics/Review.hs
arybczak: Yeah, let's remove it.

arybczak:
Also, let's put all images into a separate directory, `optics/png` or something.

phadej:
@arybczak they have to be in subdirectory of package, otherwise it becomes messy.

adamgundry:
@arybczak I went with your colour suggestions and moved the existing diagrams to `optics/diagrams` and the new ones to `optics-core/diagrams`.

I also took the liberty of copying `reoptics.png` into `optics-core/diagrams` so it can be included in the `Optics.Re` haddocks.

Makefile
```diff
@@ -39,7 +39,11 @@ codegen-join :
 
 DIAGRAMS=optics reoptics indexedoptics
 
-diagrams : optics/optics.png optics/reoptics.png optics/indexedoptics.png
+diagrams : $(DIAGRAMS:%=optics/diagrams/%.png)
```
phadej: `reoptics` is now in two places, but is second manually updated?

PR #211, opened 2019-07-03
Add at' that works for all At instances

arybczak:
I had a moment of enlightenment 🧠

Whether using `at'` with `view` should force the value is debatable.

### DIFF
--- a/optics-core/src/Optics/At/Core.hs
+++ b/optics-core/src/Optics/At/Core.hs
@@ -29,11 +29,12 @@ module Optics.At.Core
   , IxValue
 
     -- * Ixed
-  , Ixed(ix)
+  , Ixed(..)
   , ixAt
 
     -- * At
   , At(..)
+  , at'
   , sans
 
   -- * Contains
@@ -54,6 +55,7 @@ import Data.Tree
 
 import Data.Maybe.Optics
 import Optics.AffineTraversal
+import Optics.Iso
 import Optics.Lens
 import Optics.Setter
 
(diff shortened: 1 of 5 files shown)


optics-core/src/Optics/At/Core.hs
```diff
@@ -366,10 +368,45 @@ class Ixed m => At m where
   -- >>> at 1 ?~ "hello" $ Map.empty
   -- fromList [(1,"hello")]
   --
+  -- /Note:/ Usage of this function might introduce space leaks if you're not
+  -- careful to make sure that values put inside the 'Just' constructor are
+  -- evaluated. If this is the case, use 'at'' instead.
```
adamgundry:
```suggestion
  -- evaluated. To force the values and avoid such leaks, use 'at'' instead.
```

optics-core/src/Optics/At/Core.hs
```diff
@@ -366,10 +368,45 @@ class Ixed m => At m where
...
 
+-- | Version of 'at' strict in the value inside the `Just` constructor.
+--
+-- Example:
+--
+-- >>> (at () .~ Just (errorWithoutStackTrace "oops") $ Nothing) `seq` ()
+-- ()
+--
+-- >>> (at' () .~ Just (errorWithoutStackTrace "oops") $ Nothing) `seq` ()
+-- *** Exception: oops
```
adamgundry: Is there any reason this couldn't just be `error`, and use `...` to satisfy doctest?
```
-- >>> (at' () .~ Just (error "oops") $ Nothing) `seq` ()
-- *** Exception: oops
-- ...
```

arybczak:
> Interestingly this is stricter than _1' (should that be the case?)

`_1'` only forces the outer constructor, you force all fields - using `_1''` for accessing a tuple essentially makes it a strict tuple.

I guess that's the other problem of strict function variants heh - they introduce varying degree of strictness.

optics-core/src/Optics/At/Core.hs
arybczak: Oh, I didn't know that trick. Yes, doing that makes sense.

Issue #196, opened 2019-06-16
Make positional selectors/at stricter by default, provide lazy versions


arybczak:
Currently `_1`, `_2` etc. for tuples use irrefutable patterns and there are `_1'`, `_2'` variations for stricter version. I'd argue that unless you do something convoluted, you don't really need such amount of laziness.

Also, these lazy variants seem to be causing space leaks (as per https://github.com/ekmett/lens/issues/83#issuecomment-71064038).

Same thing happens with `at` (https://github.com/ekmett/lens/issues/574)

I think it's better to have these stricter and provide lazy versions for these who need them.

Also, recognizing things which are too strict is immediate (you get either a loop or thrown exception), but that doesn't hold true for space leaks.

I think we want `_1` and `at` strict and to have `_1L` and `atL` variants for lazy versions (naming up to discussion).

arybczak:
The problem that popped up is that while it's possible to strictify lazy versions, it's not possible to "lazify" strict versions. I'd therefore opt for leaving positional selectors as-is. I'm not sure what to do with `at` though. @adamgundry what do you think?

adamgundry:
I don't understand the problem. Why wouldn't we define both the lazy and strict versions?

Ideally I think we'd have both strict and lazy versions of `at` for any instance of `At`. I'm not sure if that needs both methods in the class though, or if we can have the lazy one as a class method and derive the strict one?

arybczak:
> I don't understand the problem. Why wouldn't we define both the lazy and strict versions?

There is no problem per se, it's just a ton of instances (I'm talking about positional selectors, not `at`).

`at` problem is solved in #211, but I'm still not sure about making stricter versions the default - it does go against all the conventions...

I'm leaning towards just leaving it as is.

adamgundry:
I don't feel really strongly either way about strict-by-default, so leaving the default alone is okay with me.

On a somewhat related note, I notice that the haddocks for `generateLazyPatterns` say this:
> When using lazy optics the strict optic can be recovered by composing with $!:
> ```
> strictOptic = ($!) . lazyOptic
> ```
This is of course false in `optics`, though it (apparently) holds in `lens`. Should we offer such a combinator? Or just delete the comment?

arybczak:
Oh, interesting. I tried to make an `Iso` that would be equivalent to this, but I don't think it's possible with `Iso`.

However, if you wrap ($!) in `lensVL`, then I it should work. Yes, we should include that.

PR #214, opened 2019-07-03
Add equality'

arybczak:
For https://github.com/well-typed/optics/issues/196#issuecomment-508238418


optics-core/src/Optics/Lens.hs
```diff
@@ -174,6 +175,17 @@ withLensVL o k = k (toLensVL o)
 ----------------------------------------
 -- Lenses
 
+-- | Strict version of 'equality'.
```
adamgundry: To make the Haddock link work:
```suggestion
-- | Strict version of 'Optics.Iso.equality'.
```

PR #218, opened 2019-07-08
Make construction from VL optics more visible

arybczak:
If `*VL` functions are not in the `Introduction` section, but at the end in the `van Laarhoven` section, people will have hard time discovering them, so I moved them up.

### DIFF
--- a/optics-core/src/Optics/AffineTraversal.hs
+++ b/optics-core/src/Optics/AffineTraversal.hs
@@ -16,6 +16,7 @@ module Optics.AffineTraversal
 
   -- * Introduction
   , atraversal
+  , atraversalVL
 
   -- * Elimination
   -- | An 'AffineTraversal' is a 'Optics.Setter.Setter', therefore you can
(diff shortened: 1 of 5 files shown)


adamgundry:
Hmm, I've wondered about this before, but I'm not entirely sure about this arrangement. I see what you mean about discoverability, but on the other hand it's a bit odd to have the "canonical" introduction forms include these functions without any reference in the elimination/computation/well-formedness sections. It is also a shame not to have the `*VL` introduction form as part of the `van Laarhoven` section.

Perhaps we could restore the previous position, but include a comment higher up the module as a forward pointer? For example the lens introduction section could include a comment saying something like "You can construct a `Lens` from the van Laarhoven representation using `lensVL`." We could possibly do something similar for the elimination forms.

It would also be good to point to `optics-vl` from the pages for prisms/isos, which we can do in the same way, albeit without nice Haddock links.

arybczak:
I've made #220 for this :+1:

PR #219, opened 2019-07-09
A few improvements

arybczak:
I renamed a few indexed functions to fit the `i*` pattern, improved some documentation and made `devoid` an indexed lens, because why not :thinking:


adamgundry:
Thanks for this, especially for the documentation improvements. :smile:

PR #222, opened 2019-08-14
Build optics-core with GHCJS

phadej:


.travis.yml
```diff
@@ -137,64 +163,80 @@ install:
...
-  - (cd optics-extra-* && ${CABAL} -vnormal check)
-  - (cd optics-sop-* && ${CABAL} -vnormal check)
-  - (cd optics-th-* && ${CABAL} -vnormal check)
-  - (cd optics-vl-* && ${CABAL} -vnormal check)
-  - (cd generic-optics-* && ${CABAL} -vnormal check)
-  - (cd template-haskell-optics-* && ${CABAL} -vnormal check)
-  - (cd metametapost-* && ${CABAL} -vnormal check)
+  - if ! $GHCJS && [ $HCNUMVER -lt 80802 ] ; then (cd ${PKGDIR_optics} && ${CABAL} -vnormal check) ; fi
+  - (cd ${PKGDIR_optics_core} && ${CABAL} -vnormal check)
+  - if ! $GHCJS && [ $HCNUMVER -lt 80802 ] ; then (cd ${PKGDIR_optics_extra} && ${CABAL} -vnormal check) ; fi
```
phadej: hmm. let me fix this uglyness.

we don't need upper bound, as we don't test with GHC HEAD, should we?

arybczak: Compilation with 8.8.1 used to work: https://travis-ci.org/well-typed/optics/builds/565907397

phadej:
8.8.1 is green now.

.travis.yml
```diff
@@ -137,64 +161,79 @@ install:
...
+  - ${CABAL} v2-build $WITHCOMPILER --disable-tests --disable-benchmarks all | color_cabal_output
   # Building with tests and benchmarks...
   # build & run tests, build benchmarks
-  - ${CABAL} v2-build -w ${HC} ${TEST} ${BENCH} all | color_cabal_output
+  - ${CABAL} v2-build $WITHCOMPILER ${TEST} ${BENCH} all | color_cabal_output
   # Testing...
-  - ${CABAL} v2-test -w ${HC} ${TEST} ${BENCH} all | color_cabal_output
+  - if ! $GHCJS ; then ${CABAL} v2-test $WITHCOMPILER ${TEST} ${BENCH} all | color_cabal_output ; fi
   # Doctest...
-  - "doctest --fast $(find */src  -name '*.hs' ! -path 'template-haskell-optics*/*') -XBangPatterns -XDefaultSignatures -XFlexibleContexts -XFlexibleInstances -XFunctionalDependencies -XDeriveFunctor -XGADTs -XLambdaCase -XMultiParamTypeClasses -XRankNTypes -XScopedTypeVariables -XTupleSections -XTypeFamilies -XTypeOperators -XDataKinds -XTypeApplications -XInstanceSigs"
```
arybczak: Why not test everything in one go?

phadej: I really don't see a good reason to do it in one go. I doubt it would be noticeably faster, e.g.

I.e. I see it as a very good thing, that we can `doctest` each package in separation.

.travis.yml
```diff
@@ -40,15 +41,36 @@ matrix:
...
+  allow_failures:
+    - compiler: ghc-8.8.1
 before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" = "linux" ]; then
+      sudo add-apt-repository -y ppa:hvr/ghcjs;
+      curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | sudo apt-key add -
+      sudo apt-add-repository 'https://deb.nodesource.com/node_8.x xenial main'
+      sudo apt-get update;
+      sudo apt-get install $CC cabal-install-3.0 nodejs;
```
adamgundry: Shouldn't some of this be conditional on `$GHCJS`? It seems to be with the latest `haskell-ci`.

.travis.yml
```diff
@@ -40,15 +41,36 @@ matrix:
       addons: {"apt":{"sources":["hvr-ghc"],"packages":["ghc-8.2.2","cabal-install-2.4"]}}
     - compiler: ghc-8.0.2
       addons: {"apt":{"sources":["hvr-ghc"],"packages":["ghc-8.0.2","cabal-install-2.4"]}}
+  allow_failures:
+    - compiler: ghc-8.8.1
```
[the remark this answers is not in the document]
phadej:
```
% cabal new-build -w ghc-8.8.1 all
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: generic-optics-0.1 (user goal)
[__1] trying: profunctors-5.4 (dependency of generic-optics)
[__2] trying: tagged-0.8.6 (dependency of profunctors)
[__3] trying: template-haskell-2.15.0.0/installed-2.1... (dependency of
tagged)
[__4] trying: base-4.13.0.0/installed-4.1... (dependency of generic-optics)
[__5] trying: metametapost-0.1 (user goal)
[__6] next goal: generics-sop (dependency of metametapost)
[__6] rejecting: generics-sop-0.5.0.0, generics-sop-0.4.0.1,
generics-sop-0.3.2.0 (conflict: template-haskell==2.15.0.0/installed-2.1...,
generics-sop => template-haskell>=2.8 && <2.15)
...
[__6] fail (backjumping, conflict set: base, generics-sop, metametapost,
template-haskell)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: template-haskell, tagged, base,
profunctors, generics-sop, generic-optics, metametapost
```

I.e. without relaxing bounds (and possibly `head.hackage`), full optics suite cannot be built.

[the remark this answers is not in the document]
phadej: that to say: if one tests with extensive `--allow-newer` and/or head.hackage patches, it's testing against unreleased or unsupported configurations. And in fact, it does fail randomly.

.travis.yml
phadej: Yes. Fixed.

.travis.yml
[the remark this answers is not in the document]
adamgundry: Okay, I guess we might indeed end up with false failures; it would just be a shame if we accidentally broke 8.8 support ourselves. But I guess that is relatively unlikely, at least in the short term, and we can revisit this when the ecosystem has caught up a bit.

.travis.yml
arybczak: Sounds good. I have no opinion about it anyway, was just curious about the reasoning :+1:

arybczak:
Thanks for the PR. Is it good to merge it now?

arybczak:
Also, what subpackages don't compile with GHCJS (as you're testing only optics-core)?

adamgundry:
Let's just merge this; we can always investigate GHCJS support of other subpackages later.

PR #225, opened 2019-08-18
Do not re-export Optics.Optic from optic kind modules

adamgundry:
This stops `Optics.Lens` and other optic kind modules from re-exporting `Optics.Optic`. Instead it (or `Optics.Core` or `Optics`) must be imported explicitly.

I'm in two minds about this change. The motivation is that it makes the GHCi commands like `:browse Optics.Lens` more useful; at the moment the output is swamped by all the things re-exported from `Optics.Optic`. On the other hand it will typically mean an extra import if you only want to import a few optic kinds.

Opinions?

### DIFF
--- a/optics-core/src/Data/IntMap/Optics.hs
+++ b/optics-core/src/Data/IntMap/Optics.hs
@@ -47,6 +47,7 @@ import Data.IntMap as IntMap
 
 import Optics.IxAffineTraversal
 import Optics.IxFold
+import Optics.Optic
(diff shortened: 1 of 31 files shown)


arybczak:
I agree, `:browse` output looks much better with this PR :+1:

adamgundry:
I'm going to be bold and merge this; adding the re-exports back seems easier than taking them away after a release.

PR #226, opened 2019-08-21
Better documentation for composition of indexed optics

adamgundry:
Thanks to @kosmikus for the feedback.

This introduces some duplicate export warnings, but they should be fixed by #225.


adamgundry:
Rebase, and the warnings are gone!

PR #229, opened 2019-08-30
Cabal file tweaks

adamgundry:
After uploading package candidates I spotted a few `.cabal` metadata tweaks needed.


adamgundry:
Updated to cover `optics-th` as well, which I forgot earlier.

PR #231, opened 2019-09-02
Relax optics-th bound on template-haskell

adamgundry:
I spotted after releasing 0.1 (thanks to the matrix builder) that `optics-th` has an unnecessarily tight bound on `template-haskell`, which rules out GHC 8.8. This wasn't spotted before because our CI for 8.8 is still using `allow-newer`. I've made a Hackage revision of `optics-th` to relax this bound.


adamgundry:
CI for 8.8 is currently in flux, but both my local machine and the matrix builder like this patch, so I'm going to merge anyway.

PR #235, opened 2019-09-03
Fix minor mistakes in haddocks

jhrcek:
I'm very excited about optics packages. Thank you for the wonderfully documentation. In my world you've just won the "documentation of the year" award :laughing:  I especially like the idea of structuring the docs around formation/introduction/elimination and the subtyping diagrams, which will make it very easy for me to master the APIs. I wish more Haskell packages had docs like that.. 

May I ask what you're using to generate those hierarchy images?
EDIT: nevermind, I found the answer in the readme.

### DIFF
--- a/optics-core/src/Optics/IxSetter.hs
+++ b/optics-core/src/Optics/IxSetter.hs
@@ -1,9 +1,9 @@
 {-# LANGUAGE DataKinds #-}
 -- |
 -- Module: Optics.IxSetter
--- Description: An indexed version of an 'Optics.Setter.Setter'.
+-- Description: An indexed version of a 'Optics.Setter.Setter'.
 --
--- An 'IxSetter' is an indexed version of an 'Optics.Setter.Setter'. See the
+-- An 'IxSetter' is an indexed version of a 'Optics.Setter.Setter'. See the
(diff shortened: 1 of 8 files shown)


jhrcek:
I'll probably fix more things as I find them while reading haddocks and them remove the WIP from the title when I'm done..

optics-core/src/Optics/IxSetter.hs
```diff
@@ -1,9 +1,9 @@
 {-# LANGUAGE DataKinds #-}
 -- |
 -- Module: Optics.IxSetter
--- Description: An indexed version of an 'Optics.Setter.Setter'.
+-- Description: An indexed version of a 'Optics.Setter.Setter'.
```
jhrcek: Most of these article changes optimize for how it reads in the rendered haddock (which doesn't show the module name prefix). E.g.
Screenshot from 2019-09-03 14-51-30

adamgundry:
Thanks for the corrections, much appreciated!

> May I ask what you're using to generate those hierarchy images?
> EDIT: nevermind, I found the answer in the readme.

Just for completeness: the `metametapost` library in this repo defines a Haskell EDSL for MetaPost, which we use to turn a description of the hierarchy into a variety of images.

PR #239, opened 2019-09-04
Add non, non' and anon

### LINKED ISSUE #238: Is there an equivalent to 'non' from lens?
Thanks for this library, I'm really excited about it! I'm just starting to explore and tentatively see if I can convert some of my `lens`-using code to `optics`.

There's one place where I use [non](https://hackage.haskell.org/package/lens-4.17.1/docs/Control-Lens-Iso.html#v:non) with a `Map`, just like the examples in the lens documentation. I'm not finding any corresponding `Iso` in optics. I imagine it would look something like this:

```haskell
non :: Eq a => a -> Iso' (Maybe a) a
non z = iso f g
  where
    g x | x == z     =  Nothing
        | otherwise  =  Just x

    f Nothing   =  z
    f (Just x)  =  x
```

Was this an intentional omission because `non` (as noted in the lens docs) is not a true isomorphism? Is there another way to achieve the same effect with optics?

arybczak:
Fixes #238 - I'm not sure if we want to have `non'` and `anon` though, they feel a bit weird to me. Any opinions?

### DIFF
--- a/optics-core/CHANGELOG.md
+++ b/optics-core/CHANGELOG.md
@@ -1,0 +1,2 @@
+# optics-core-0.2 (????-??-??)
+* Add non, non' and anon to Optics.Iso
--- a/optics-core/src/Optics/Iso.hs
+++ b/optics-core/src/Optics/Iso.hs
@@ -52,6 +52,9 @@ module Optics.Iso
   , coerced
   , coercedTo
   , coerced1
+  , non
+  , non'
+  , anon
   , curried
   , uncurried
   , flipped
@@ -80,7 +83,11 @@ module Optics.Iso
 import Data.Tuple
 import Data.Bifunctor
 import Data.Coerce
+import Data.Maybe
 
+import Optics.AffineFold
+import Optics.Prism
+import Optics.Review
 import Optics.Internal.Concrete
 import Optics.Internal.Optic
 import Optics.Internal.Profunctor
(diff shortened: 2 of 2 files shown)


arybczak:
Ok, so

```haskell
non = non' . only
anon = non' . nearly
```

Maybe it makes sense to only include `non'` (and rename it to `non`) then?

chris-martin:
Would it make sense to also add the documentation regarding the `Map` use cases to `Data.Map.Optics` to aid discoverability?

chris-martin:
> Maybe it makes sense to only include `non'` (and rename it to `non`) then?

That seems sensible to me.

Practically, given the one use case I know: instead of writing `non mempty`, one would write `non _Empty`, which is about the same.

Pedagogically, I like reducing the number of functions, or at least emphasizing which one is the most "fundamental" of them (which the name `non'` fails to do), and I really appreciate a function that converts a `Prism` to an `Iso` because it lets me keep my thoughts contained within the optics family.

adamgundry:
> Maybe it makes sense to only include `non'` (and rename it to `non`) then?

In cases like this I'd be inclined to retain the `lens` terminology rather than introducing small differences in pursuit of some minimal subset. Of course we won't be perfectly compatible with `lens`, but having `non`, `non'` and `anon` feels reasonable to me.

arybczak:
> Would it make sense to also add the documentation regarding the Map use cases to Data.Map.Optics to aid discoverability?

I think it would. If you have something specific in mind, please make a PR :)

I'm gonna merge this if there are no objections.

arybczak:
I added a changelog.

PR #248, opened 2019-09-06
Introduce type synonyms for optic kind and indices

### LINKED ISSUE #71: Check is type OpticKind = Type may improve haddocks
EDIT: Also Add `IxList = [Type]`

Both require enabling `TypeInType` (to be used in kind annotation); but that should be ok.

adamgundry:
Fixes #71. This improves the haddocks and repl use: 
```hs
*Optics.Core> :kind Optic
Optic :: OpticKind -> Indices -> * -> * -> * -> * -> *
```
We need to enable `TypeInType` for older GHCs. This has an effect on the parsing of `*`, so I've switched to `Type` in a few places.

Opinions on `Indices` vs `IxList`?

### DIFF
--- a/optics-core/src/Optics/Internal/Indexed.hs
+++ b/optics-core/src/Optics/Internal/Indexed.hs
@@ -38,7 +38,7 @@ import Optics.Internal.Profunctor
 import Optics.Internal.Utils
 
 -- | Show useful error message when a function expects optics without indices.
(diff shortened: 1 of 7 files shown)


arybczak:
I'm for `IxList` as it says that it's a list ;)

arybczak:
Looks good :+1:

PR #253, opened 2019-09-06
More testing of compilation with GHCJS

arybczak:


arybczak:
Everything (apart from metametapost for some reason) compiles with GHCJS now, it should be good to merge.

PR #254, opened 2019-09-06
Add combinators cheat sheet

### LINKED ISSUE #241: Port the operators cheat sheet from lens
https://github.com/ekmett/lens/wiki/operators

adamgundry:
Fixes #241.


phadej:
This looks good. Should we include https://hackage.haskell.org/package/optics-extra-0.1/docs/Optics-State-Operators.html operators (and `Optics.State` functions) in the table?

optics/src/Optics.hs
```diff
@@ -847,6 +850,94 @@ import Data.Either.Optics                    as P
...
+-- +--------------------+---------------------------------------------------+
+-- | 'traverseOf'       | Update target(s) with an 'Applicative'.           |
+-- +--------------------+---------------------------------------------------+
+-- |              __'Prism's__                                              |
+-- +--------------------+---------------------------------------------------+
+-- | 'prism'            | Build a 'Prism' from a constructor and matcher.   |
+-- +--------------------+---------------------------------------------------+
+-- | 'review' / '#'     | Use a 'Prism' to construct the sum type.          |
+-- +--------------------+---------------------------------------------------+
+-- | 'preview' / '^?'   | Match the target or return 'Nothing'.             |
```
arybczak: It might be good to create a new section for `AffineFold` and move preview/previews there (same for `IxAffineFold` for ipreview/ipreviews).

optics/src/Optics.hs
```diff
@@ -858,85 +858,112 @@ import Data.Either.Optics                    as P
...
+-- +-----------------------------------+---------------------------------------------------+
+-- |'sets'                             |Build a 'Setter' from an update function.          |
+-- +-----------------------------------+---------------------------------------------------+
+-- |'mapped'                           |Build a 'Setter' from the 'Functor' class.         |
+-- +-----------------------------------+---------------------------------------------------+
+-- |'set' / '.~'                       |Replace target(s) with value.                      |
+-- +-----------------------------------+---------------------------------------------------+
+-- |'over' / '%~'                      |Replace target(s) by applying a function.          |
+-- +-----------------------------------+---------------------------------------------------+
+-- |'assign' /                         |'Control.Monad.State.MonadState' version of 'set'; |
```
phadej: i thought of making third column as in `lens` table; that would highlight the relationship between `%~` and `%=` etc. Yet, I don't know how to combine notes reasonably

adamgundry: Yes, I wondered about that but wasn't very keen on adding extra columns just for two rows to have stateful variants. Perhaps they should appear in a completely different table?

phadej: @adamgundry otoh, I actually didn't realise we don't have `use`, I didn't need it yet myself, but surely some `MonadState` / `Zoom` will ask for it. 

I.e. I'm not sure how long it will stay at only two rows :)

adamgundry: I've made a new table instead, which I think looks better as it illustrates the relationships more clearly.

optics/src/Optics.hs
adamgundry: Done, thanks.

optics/src/Optics.hs
phadej: Indeed :+1:

arybczak: > I actually didn't realise we don't have use

We do, it's in `Optics.View`.

phadej: > We do, it's in Optics.View.

ou. "this is terrible". It should be `guse` if it's there, and have proper `use`, `preuse` (someething for list), without `lens`-weirdness.

arybczak: PRs welcome ;)

adamgundry: Let's move this to #261.

PR #256, opened 2019-09-07
Split out indexed-profunctors package

### LINKED ISSUE #247: Separate library for indexed profunctor optics?
Would it be worth moving some of the "internal implementation details" of `optics-core` into a separate package for indexed profunctor optics? This has been requested at https://www.reddit.com/r/haskell/comments/cyo4o2/welltyped_announcing_the_optics_library/ez73jei/ and also in discussons with @kcsongor it might be useful for `generic-(lens|optics)`.

adamgundry:
Fixes #247. This moves most of the contents of `Optics.Internal.{Profunctor,Tagged,Concrete}`
to `Optics.Internal.Profunctor` and exposes them in a new package.

What should the new package be called? Is `indexed-profunctors` too generic?

CC @kcsongor


arybczak:
> Is indexed-profunctors too generic?

I think so, especially since the only module it has is called `Optics.Internal.Profunctor` ;) Perhaps `optics-indexed-profunctors`? Or `optics-prim`?

It looks good, but there is surprisingly little code split. Is it enough to help other packages like generic-lens?

adamgundry:
Just discussed this with @kcsongor and we're inclined to keep the package name `indexed-profunctors` but rename the module to `Data.Profunctor.Indexed` to fit with the existing `profunctors` namespace. How does that sound?

> It looks good, but there is surprisingly little code split. Is it enough to help other packages like generic-lens?

Yes, see https://github.com/kcsongor/generic-lens/tree/indexed-profunctors :smile:

arybczak:
> How does that sound?

I guess it's alright, at least until someone else has the idea of indexed profunctors that clashes with the one in this package, haha.

phadej:
I think we should have use non-Internal module name, e.g.
- `Optics.Profunctor.Indexed`
- or just `Optics.Profunctor`
- or `Data.Profunctor.Indexed`
- or `Data.Profunctor.Optics`

I don't have a favourite

adamgundry:
I've renamed the module to `Data.Profunctor.Indexed` and uploaded https://hackage.haskell.org/package/indexed-profunctors-0.1/candidate. Any objections to merging and releasing so we can unblock @kcsongor?

phadej:
@adamgundry :+1:

Should we make `optics-0.2` release too (incl.currently opened PRs, which don't have anything controversial, #265 and #266 )

adamgundry:
> Should we make `optics-0.2` release too

Yes, I'm hoping to get to this soon.

PR #261, opened 2019-09-15
Make use like view, without ViewResult

phadej:


arybczak:
How about adding a version for traversals called `multiuse`?

phadej:
I think that `asks $ toListOf someTraversal` (or `gets`) is not that bad.

I wouldn't hold this going in just because there is no good name.

phadej:
For the record, I'm against editing changelogs in PRs. They conflict.

phadej:
And looks like there isn't Changelog file in `optics-extra`, so the one making release, please remember this PR. 

cc @adamgundry

arybczak:
> please remember this PR

Yeah, that's why changelog changes need to be made in PRs 😞

phadej:
@arybczak if you are willing to resolve conflicts in my (and others') PR, then we can write that (i.e. add entry to changelog, **we** will resolve conflicts before merging) into `CONTRIBUTING.md`

arybczak:
Fair enough ;)

PR #262, opened 2019-09-15
Kinded ixed

phadej:


optics-core/src/Optics/At/Core.hs
```diff
@@ -115,7 +118,13 @@ instance Ord a => Contains (Set a) where
...
 -- values stored in the container, for example @'IxValue' ('Map' k a) ~ a@. This
 -- is shared by both 'Ixed' and 'At'.
-type family IxValue (m :: *) :: *
+type family IxValue (m :: Type) :: Type
+
+-- | Type family that takes a key-value container type and returns the kind
+-- of optic to index into it. For most containers, it's 'An_AffineTraversal',
+-- @Representable@ (Naperian) containers it is 'A_Lens', and multi-maps would
+-- have 'A_Traversal'.
+type family IxKind (m :: Type) :: OpticKind
```
adamgundry: Would it make sense to have `IxKind` be an associated type in `Ixed`, rather than a top-level type family? Then it could have a default definition of `An_AffineTraversal`, which would save a few LOC and be more backwards/`lens`-compatible.

optics-core/cabal.project
```diff
@@ -0,0 +1 @@
+packages: .
```
adamgundry: I don't think this should have snuck in?

arybczak:
PR looks good to me, I'd just prefer the kind to be a fundep, it'll be clearer I think.

phadej:
@arybczak, @adamgundry cleanest approaches would be to have either everything by fundeps or as associated type-families. Yet, having only an `IxKind` as associated family is "smallest change" for now., good idea @adamgundry. I have done that.

phadej:
Having a slightly stronger type for Naperian containers is hard. I failed to come with name for last two years.

arybczak:
You removed optics-core.cabal.

phadej:
@arybczak luckily we have CI and reviewers!

phadej:
Travis is happy.

PR #265, opened 2019-10-02
Add simpleFieldLabels and simpleNamer

arybczak:


arybczak:
I'm not sure if `simple{FieldLabels,Namer}` is a good name. Perhaps `noPrefix{FieldLabels,Namer}` would be better?

phadej:
@adamgundry as a native speaker could comment :) I'd go with something like `directNamer`, `intactNamer` or `pristineNamer` (again, I'm not 100% about nuance differences these might have).

arybczak:
@adamgundry ping :)

adamgundry:
Sorry for being slow (on this and other things), I've just been snowed under.

Given that we have `underscoreNoPrefixNamer` and `classUnderscoreNoPrefixNamer` I'm somewhat inclined towards `noPrefixNamer`/`noPrefixFields`. I think various synonyms for "unchanging" work in a `...Namer` context (`identityNamer`?), but for `...Fields` something like `identityFields` or `pristineFields` seems more confusing as it doesn't clearly refer to the names.

Presumably this doesn't generate top-level lens definitions with the same name as the selector functions, though? How does that work?

arybczak:
> Presumably this doesn't generate top-level lens definitions with the same name as the selector functions, though? How does that work?

It doesn't, i.e. if you try to use it with `makeLensesWith`, you'll get compilation errors (multiple declarations of ...) (`declateLensesWith` works). It's supposed to be used for generation of OpticLabel instances.

arybczak:
I changed `simple` to `noPrefix`.

PR #266, opened 2019-10-10
Correct documentation for viewing type-changing optics

### LINKED ISSUE #263: view for Iso
```haskell
view :: Iso s t a b -> s -> a
```

This seems to be false.

The actual definition of `view` is:

```haskell
Is k A_Getter => Optic' k is s a -> s -> a
```

which in our case is essentially:

```haskell
Is k A_Getter => Iso' s a -> s -> a
```

... which is pretty sad...

adamgundry:
Fixes #263.

@ilyakooo0 thanks for flagging up the documentation errors here. Does this look better to you?

### DIFF
--- a/optics-core/src/Optics/Getter.hs
+++ b/optics-core/src/Optics/Getter.hs
@@ -47,6 +47,10 @@ import Optics.Internal.Profunctor
 type Getter s a = Optic' A_Getter NoIx s a
 
 -- | View the value pointed to by a getter.
+--
+-- If you want to 'view' a type-modifying optic that is insufficiently
+-- polymorphic to be type-preserving, use 'Optics.ReadOnly.getting'.
+--
 view :: Is k A_Getter => Optic' k is s a -> s -> a
 view o = views o id
 {-# INLINE view #-}
--- a/optics-core/src/Optics/Prism.hs
+++ b/optics-core/src/Optics/Prism.hs
@@ -20,13 +20,21 @@ module Optics.Prism
   -- specialise types to obtain:
   --
   -- @
-  -- 'Optics.AffineFold.preview' :: 'Prism' s t a b -> s -> Maybe a
-  -- 'Optics.Review.review'  :: 'Prism' s t a b -> b -> t
+  -- 'Optics.AffineFold.preview' :: 'Prism'' s a -> s -> Maybe a
+  -- 'Optics.Review.review'  :: 'Prism'' s a -> a -> s
   -- @
   --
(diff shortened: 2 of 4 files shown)


optics-core/src/Optics/Prism.hs
```diff
@@ -20,14 +20,21 @@ module Optics.Prism
   -- specialise types to obtain:
   --
   -- @
-  -- 'Optics.AffineFold.preview' :: 'Prism' s t a b -> s -> Maybe a
-  -- 'Optics.Review.review'  :: 'Prism' s t a b -> b -> t
+  -- 'Optics.AffineFold.preview' :: 'Prism'' s a -> s -> Maybe a
+  -- 'Optics.Review.review'  :: 'Prism'' s a -> a -> s
```
phadej: I'm quite sure `::` should be aligned

optics-core/src/Optics/Getter.hs
```diff
@@ -47,6 +47,9 @@ import Optics.Internal.Profunctor
 type Getter s a = Optic' A_Getter NoIx s a
 
 -- | View the value pointed to by a getter.
+--
+-- If you want to 'view' a type-changing optic, you may need to use
```
phadej: On the second thought: *a type-changing optic* doesn't sound right.

I'm not sure whether *an insufficiently polymorphic type-changing optic* would be any better, but I think less wrong?

optics-core/src/Optics/Prism.hs
adamgundry: They are aligned in the rendered output (because the module prefix isn't displayed).

phadej: I see, thanks for the explanation.

optics-core/src/Optics/Getter.hs
adamgundry: Yes, I wondered what to call these. Your suggestion is a bit better, and I've reworded it slightly further.

optics-core/src/Optics/Getter.hs
```diff
@@ -47,6 +47,10 @@ import Optics.Internal.Profunctor
 type Getter s a = Optic' A_Getter NoIx s a
 
 -- | View the value pointed to by a getter.
+--
+-- If you want to 'view' a type-modifying optic that is insufficiently
+-- polymorphic to be type-preserving, use 'Optics.ReadOnly.getting'.
```
phadej: :+1:

PR #268, opened 2019-10-17
WIP&RFC: NonEmpty traversals and folds

phadej:
At the moment, these are not so useful, as the only common
non-empty structures are `NonEmpty` and `Tree`.

After https://github.com/haskell/containers/pull/616 (nonempty maps and
sets) is done, there might be bigger interest in non-empty
traversals/folds.

This commit is a PoC that we can these relatively easily provide them.

- We need Apply, Foldable1, Traversable1 from semigroupoids
- but we can vendor them
    * Foldable1/Traversable1 are used for folded1 / traversed1,
      which we can write instances for common things: (NonEmpty, NESet, NEMap)
    * Apply is virtually hidden, if higher abstraction eliminators are used

Also I think that the interesting&useful bit is non empty Fold.
There aren't that many interesting Apply's which aren't Applicative,
but extracting NonEmpty / NESet / ... from a structure sounds useful
(e.g. maximumBy!)

There's also a https://gitlab.haskell.org/ghc/ghc/issues/13573
about adding Foldable1 to base, which can actually happen,
as it doesn't shake the existing type-class hierarchy.

### DIFF
--- a/codegen/Subtypes.hs
+++ b/codegen/Subtypes.hs
@@ -32,6 +32,8 @@ data OpticKind
   |  An_AffineTraversal
   -- | Tag for a traversal.
   |  A_Traversal
+  -- | Tag for a non empty traversal
+  |  A_Traversal1
   -- | Tag for a setter.
   |  A_Setter
   -- | Tag for a reversed prism.
@@ -40,6 +42,8 @@ data OpticKind
   |  A_Getter
   -- | Tag for an affine fold.
   |  An_AffineFold
+  -- | Tag for a nonempty fold.
+  |  A_Fold1
   -- | Tag for a fold.
   |  A_Fold
   -- | Tag for a reversed lens.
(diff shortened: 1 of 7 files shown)


arybczak:
Hmm, the main issue is that you're using `Apply` etc. incompatible with what's in semigroupoids. And they appear a lot in exposed type signatures from what I can see.

phadej:
> they appear a lot in exposed type signatures from what I can see.

Only in the named `Traversal1` module, which I argue only few people would ever use.

There just aren't that many interesting `Apply`s, yet not `Applicative`s. Probably it's your own, and then writing an instance is not in issue.

This PR is about `Fold1` (or `NonEmptyFold` or whatever one would like to call it).

arybczak:
Can we close it and resurrect (parts of) it when feasible?

PR #274, opened 2019-11-12
Add (^?!)

arybczak:
I found that this is useful in tests when you expect the value of an affine fold to be there.

It's more succinct to use `^?!` than `^?` with `fromJust`.

### DIFF
--- a/optics-core/src/Optics/Operators.hs
+++ b/optics-core/src/Optics/Operators.hs
@@ -50,7 +50,7 @@ infixl 8 ^?
 
 infixl 8 ^..
 
--- | Flipped infix version of 'review'.
+-- | Infix version of 'review'.
 (#) :: Is k A_Review => Optic' k is t b -> b -> t
 (#) = review
 {-# INLINE (#) #-}
--- a/optics-core/src/Optics/Operators/Unsafe.hs
+++ b/optics-core/src/Optics/Operators/Unsafe.hs
@@ -1,0 +1,37 @@
+-- |
+-- Module: Optics.Operators.Unsafe
+-- Description: Definitions of unsafe infix operators for optics.
+--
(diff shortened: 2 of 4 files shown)


optics-core/src/Optics/Operators.hs
```diff
@@ -43,14 +47,33 @@ infixl 8 ^.
...
+-- >>> "world" ^?! ix 3
+-- 'l'
+--
+-- >>> [] ^?! _head
+-- *** Exception: (^?!): empty affine fold
+-- ...
+--
+-- @since 0.2.1
+(^?!) :: (HasCallStack, Is k An_AffineFold) => s -> Optic' k is s a -> a
+s ^?! o = fromMaybe (error "(^?!): empty affine fold") (s ^? o)
```
phadej: It calls `error`, so IMHO should be in some `Unsafe` module.

arybczak: Makes sense, I moved it to `Optics.Operators.Unsafe`.

optics-core/src/Optics/Operators/Unsafe.hs
```diff
@@ -0,0 +1,37 @@
...
+-- Module: Optics.Operators.Unsafe
+-- Description: Definitions of unsafe infix operators for optics.
+--
+module Optics.Operators.Unsafe
+  ( (^?!)
+  )
+  where
+
+import Data.Maybe
+import GHC.Stack
```
phadej: We should explicit import lists for all external imports.

arybczak: We generally don't do this in other modules though.

phadej: We should there as well.

Issue #273, opened 2019-11-11
singular examples


phadej:
A dump of ideas:

```
19:40  i0_work [] & ix 3 %%~ (\k -> (0 :: Int, k))
19:41  i0_work • No instance for (Monoid Int) arising from a use of ‘ix’
```

The fact we have `ix` which is (in this case) `AffineTraversal`, doesn't require `singular`.

"Unfortunately" we don't have `(%%~)`, so same result doesn't work
Should we have

```haskell
class ToVL (k :: OpticKind) where
  type family VLConstraint k f :: Constraint

  toVL :: VLConstraint k f => Optic k NoIx s t a b -> (a -> f b) -> s -> f t

instance ToVL A_Lens
instance ToVL A_Traversal
```

in `optics-core`, and more complete one one in `optics-vl`?

arybczak:
We don't have `singular` at the moment though. I think of it as a generalization of `fromJust`, somewhat related to `^?!`.

As for the second part, it needs a separate issue. It was discussed a bit in #105 though.

phadej:
To avoid confusion: I think in `optics`

```haskell
-- or should this be named `affine` ?
singular :: Traversal s t a a -> AffineTraversal s t a a -- safe, focus on first element
```

where

```haskell
unsafeLens :: AffineTraversal s t a b -> Lens s t a b
unsafePrism :: AffineTraversal s t a b -> Prism s t a b
```

are something different

arybczak:
I see. Interesting question - `singular` is different in `lens` which might be confusing, but we did somewhat similar thing to `filtered`...

arybczak:
Should this be closed after #276 was merged?

phadej:
Yes, there's #275 for the second part.

PR #276, opened 2019-11-12
Add (i)singular

arybczak:
Related to #273.


optics-core/src/Optics/Fold.hs
```diff
@@ -231,6 +231,8 @@ unfolded step = foldVL $ \f -> fix $ \loop b ->
 
 -- | Convert a fold to an 'AffineFold' that visits the first element of the
 -- original fold.
+--
+-- For the traversal version see 'Optics.Traversal.singular'.
```
phadej: I didn't realize that, nice.

PR #277, opened 2019-11-12
Use stricter uncurry for better performance

arybczak:
I noticed when comparing `isingular` to `ipre` that the latter was 15% slower, tracked it down to the fact that `iafolding` that `ipre` uses was using `uncurry` whereas `isingular` has regular pattern matching on a pair.


arybczak:
I'll merge on Monday unless anyone disagrees.

Issue #245, opened 2019-09-05
Export Curry/CurryCompose or hide icomposeN?


adamgundry:
The inferred type of `icomposeN` exposes internals:
```hs
*Optics> :t icomposeN
icomposeN
  :: (Optics.Internal.Optic.TypeLevel.CurryCompose is,
      NonEmptyIndices is) =>
     Optics.Internal.Optic.TypeLevel.Curry is i
     -> Optic k is s t a b -> Optic k (WithIx i) s t a b
```
I think we should either export these from `Optics`, or perhaps not expose `icomposeN`...

arybczak:
> I think we should either export these from Optics

Sounds good to me :+1:

PR #278, opened 2019-11-14
Export Curry/CurryCompose

arybczak:
Fixes https://github.com/well-typed/optics/issues/245.


arybczak:
I'll merge on Monday unless anyone disagrees.

optics-core/src/Optics/Optic.hs
```diff
@@ -49,6 +49,8 @@ module Optics.Optic
   , NonEmptyIndices
   , HasSingleIndex
   , AcceptsEmptyIndices
+  , Curry
```
phadej: Does these have proper haddocks?

arybczak: Yes.

PR #280, opened 2019-11-19
Add computation QuickCheck properties

phadej:
We do have inspection tests for these,
yet QuickCheck finds counter-examples.

### DIFF
--- a/optics/optics.cabal
+++ b/optics/optics.cabal
@@ -136,9 +136,11 @@ test-suite optics-tests
                , mtl
                , optics
                , optics-core
+               , QuickCheck
                , random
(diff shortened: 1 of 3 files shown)


phadej:
ping

arybczak:
Looks good :+1: I'm not sure what you mean by

> yet QuickCheck finds counter-examples.

though since tests pass :thinking:

phadej:
QuickCheck finds counter-examples for few properties which don't hold:

```haskell
in expectFailure prop
```

Which I try to point out with `~=`, i.e. often the equality works, but not always.

PR #284, opened 2019-12-15
Test with GHC 8.10.1

arybczak:


optics-core/src/Optics/Internal/Optic/Subtyping.hs
```diff
@@ -31,7 +31,7 @@ instance {-# OVERLAPPABLE #-} TypeError ('ShowType k
                                          ':<>: 'Text " cannot be used as "
                                          ':<>: 'ShowType l
                                         ) => Is k l where
-  implies = error "unreachable"
+  implies _ _ = error "unreachable"
```
phadej: So we'll **need** (at least a patch) release to support 8.10.1. Yet, I'm positively surprised so little broke

phadej:
Is `optics` package left out on purpose?

arybczak:
Kinda, I'm adding things gradually to see what fails (if anything).

optics/optics.cabal
```diff
@@ -158,6 +158,10 @@ benchmark folds
   hs-source-dirs:   benchmarks
   ghc-options:      -Wall -threaded
 
+  -- TODO: remove when deps compile with GHC 8.10
+  if impl(ghc >= 8.10)
```
phadej: I'm pretty sure that dependencies should be in `head.hackage` repository, with patches so they do compile against 8.10

Otherwise it's an issue with `head.hackage`.

arybczak:
Interesting, label tests fail to compile.

phadej:
Locally I also get

```
src/Language/Haskell/TH/Optics.hs:1609:5: error:
    • Couldn't match type ‘Maybe Exp’ with ‘Exp’
      Expected type: Prism' Exp [Exp]
        Actual type: Prism Exp Exp [Maybe Exp] [Maybe Exp]
    • In the expression: prism' reviewer remitter
      In an equation for ‘_TupE’:
          _TupE
            = prism' reviewer remitter
            where
                reviewer = TupE
                remitter (TupE x) = Just x
                remitter _ = Nothing
     |
1609 |   = prism' reviewer remitter
     |     ^^^
```

phadej:
I pushed a small commit, which allows to test locally, when you uncomment the relevant part in the `cabal.project` https://github.com/well-typed/optics/commit/ab762821f89465e19021eff89e5f296c31e88651

phadej:
I can see now `Tests/Labels.hs` failures locally true. I hope it's not something wrong with GHC.

phadej:
It's not. I'm surprised that test works with current GHCs. One need to move

```haskell
-- Data definitions

...

makeFieldLabels ''Mammal
makePrismLabels ''Mammal
makeFieldLabels ''Fish
makePrismLabels ''Fish
makeFieldLabels ''Human
```

block to be the very first thing in the source. Due to how template-haskell intra-splice parts are compiled in order.

arybczak:
Thanks for investigation, should be good to merge now.

phadej:
So, `template-haskell-optics` is left out for now? Should we make an issue for it, so it's not so easily forgotten?

arybczak:
I made `template-haskell-optics` compile with GHC 8.10.1 :+1:

PR #287, opened 2020-01-28
Generalise mapping to work on getters and reviews

phadej:


arybczak:
Is that pretty much a read only part of `mapping` from `Optics.Iso`?

arybczak:
Perhaps it's worth making `mapping` a part of an optic-kind specific type class then?

phadej:
Hmm, good point.

```
mapping :: Iso -> Iso
mapping :: Lens -> Getter
mapping :: Getter -> Getter
mapping :: Prism -> Review
mapping :: Review -> Review
```

so you could `review (mapping _Just) [1,2,3]` e.g.

Does that sound right, @arybczak ?

arybczak:
Yes, I think so :)

phadej:
Pushed a generalised version

optics/optics.cabal
```diff
@@ -83,6 +83,7 @@ library
                     , Optics.Empty
                     , Optics.Indexed
                     , Optics.Label
+                    , Optics.Mapping
```
arybczak: This needs to be re-exported in `Optics` module.

optics-core/src/Optics/Iso.hs
```diff
@@ -150,13 +144,7 @@ under k = withIso k $ \sa bt ts -> sa . ts . bt
...
 -- Isomorphisms
 
--- | This can be used to lift any 'Iso' into an arbitrary 'Functor'.
-mapping
-  :: (Functor f, Functor g)
-  => Iso    s     t     a     b
-  -> Iso (f s) (g t) (f a) (g b)
-mapping k = withIso k $ \sa bt -> iso (fmap sa) (fmap bt)
-{-# INLINE mapping #-}
+
```
arybczak: Extra newline

optics-core/src/Optics/Iso.hs
```diff
@@ -74,13 +74,7 @@ module Optics.Iso
...
   , under
 
-  -- * Combinators
-  -- | The 'Optics.Re.re' combinator can be used to reverse an 'Iso':
-  --
-  -- @
-  -- 'Optics.Re.re' :: 'Iso' s t a b -> 'Iso' b a t s
-  -- @
-  , mapping
+
```
arybczak: Extra newline

optics-core/src/Optics/Mapping.hs
```diff
@@ -0,0 +1,89 @@
...
+-- >>> import Optics.Core
+-- >>> import Optics.Operators
+
+-- | Class for optics supporting 'mapping' through a 'Functor'.
+class MappingOptic k f g s t a b where
+  -- | Type family that maps an optic to the optic kind produced by
+  -- 'mapping' using it.
+  type MappedOptic k
+
+  -- | The 'Optics.Re.mapping' can be used to lift optic through a 'Functor'.
```
arybczak: `Optics.Re.Mapping` :)

phadej:
ping @adamgundry

optics-core/src/Optics/Iso.hs
```diff
@@ -74,14 +74,6 @@ module Optics.Iso
...
   , under
 
-  -- * Combinators
-  -- | The 'Optics.Re.re' combinator can be used to reverse an 'Iso':
-  --
-  -- @
-  -- 'Optics.Re.re' :: 'Iso' s t a b -> 'Iso' b a t s
-  -- @
-  , mapping
-
```
adamgundry: We seem to have lost some documentation here. I'd suggest keeping the mention of `mapping` but in Haddock comments, like `re`.

phadej: Fixed. I was confused, it looked like the above is "forgotten to update" documentation for `mapping`.
Resurrected it so *Combinators* section doesn't export anything, so it isnt' so confusing:

Screenshot from 2020-02-10 14-46-20

Issue #140, opened 2019-03-11
OverloadedLabels requires either manually writing instances or using TemplateHaskell


phadej:
I really like `generic-lens` base approach. It was using existing library, and was so convenient.

The inconvinience is visible from the example

https://github.com/well-typed/optics/blob/f5c678166546c333e850579558fe9f9b7133c6c7/optics-core/src/Optics/Internal/Optic.hs#L191-L231

if optics wouldn't be scary topic to begin with!

arybczak:
See https://github.com/well-typed/optics/issues/131#issuecomment-464494920, generic-lens is not usable at the moment.

adamgundry:
We're not planning to address this for the initial release, but in the future we could conceivably adopt either `generic-optics` or the proposed magical `HasField` class, so leaving this open.

adamgundry:
See kcsongor/generic-lens#108 for current thinking on this.

Issue #288, opened 2020-02-13
Add Bitraversable Traversals


cfraz89:
Porting from lens, have noticed the lack of Bitraversable optics, namely `both`.

Would be nice to add `both` and `both1` to optics or optics-extra.
`both` can be defined as:

```
both :: Bitraversable r => Traversal (r a a) (r b b) a b
both = traversalVL $ \f -> bitraverse f f
```

arybczak:
That requires dropping support for GHC 8.0.2.

See https://github.com/well-typed/optics/issues/96 for more details.

adamgundry:
I previously argued for keeping 8.0.2 support, but I think enough time has gone by that we could reasonably drop it.

That said, @cfraz89 does `each` or `chosen` work in your use case? I'm not sure how often people use `both` at types other than `(,)` or `Either`.

hasufell:
I think 8.0.2 is something that still exists reasonably in projects that have a hard time upgrading their tool chain stack. Supporting it sounds worthwhile, but shouldn't it be possible to have features that at least require ghc 8.2? I don't see a problem with that.

cfraz89:
@adamgundry Yes we were just using it on `(,)`, `each` is fine here.

PR #290, opened 2020-03-05
WIP: generic-optics-labels experiment

adamgundry:
See #140 and kcsongor/generic-lens#108. This uses nasty tricks to make it possible to extend `LabelOptic` downstream, and adds a package `generic-optics-labels` that does so using `generic-optics`. We retain the custom error message if a `Generic` instance does not exist or the orphan instance in `generic-optics-labels` has not been imported.

Basic tests seem to work, although it needs more prodding to see if the wheel falls off somewhere. It turns out that my worries about polymorphism are already an issue with `LabelOptic`: if you use a polymorphic expression such as `view #foo` without a type signature, you will hit the custom type error rather than remaining polymorphic in a `LabelOptic` constraint. But that's not entirely unreasonable, perhaps.

If we go for this approach, I think we should rename `Optics.Label` to `Optics.Label.Core` in `optics-core` and make `optics` depend on `generic-optics-labels` so it can expose an `Optics.Label` module including the instance.

### DIFF
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -30,6 +30,7 @@ module Optics.Internal.Optic
   -- * Labels
   , LabelOptic(..)
   , LabelOptic'
+  , GeneralLabelOptic(..)
   -- * Re-exports
   , module Optics.Internal.Optic.Subtyping
   , module Optics.Internal.Optic.Types
@@ -40,6 +41,7 @@ import Data.Function ((&))
 import Data.Kind (Type)
 import Data.Proxy (Proxy (..))
 import Data.Type.Equality
+import GHC.Generics (Rep)
 import GHC.OverloadedLabels
 import GHC.TypeLits
 
@@ -225,13 +227,41 @@ instance
   ) => LabelOptic name k Void0 Void0 a b where
   labelOptic = Optic id
 
+-- | Type synonym for a type-preserving optic as overloaded label.
+type LabelOptic' name k s a = LabelOptic name k s s a a
+
+-- | Implements fallback behaviour in case there is no explicit 'LabelOptic'
+-- instance. This has a catch-all incoherent instance that merely yields an
+-- error message.  However, a downstream module can give a more specific
+-- instance that uses 'Generic' to construct an optic automatically.
+--
+-- To support this, the last parameter will be instantiated to 'True' if at
+-- least one of @s@ or @t@ has a 'Generic' instance.
+class GeneralLabelOptic (name :: Symbol) k s t a b (has_rep :: Bool) where
+  -- | Used to interpret overloaded label syntax in the absence of an explicit
+  -- 'LabelOptic' instance.
+  generalLabelOptic :: Optic k NoIx s t a b
+
+-- | If no instance matches, fall back on 'GeneralLabelOptic'.
+instance {-# OVERLAPPABLE #-}
+  (LabelOptic name k s t a b,  -- Needed to satisfy functional dependecy
+  GeneralLabelOptic name k s t a b (HasRep (Rep s) (Rep t))
+  ) => LabelOptic name k s t a b where
+  labelOptic = generalLabelOptic @name @k @s @t @a @b @(HasRep (Rep s) (Rep t))
+
+-- | This type family should be called with applications of 'Rep' on both sides,
+-- and will reduce to 'True' if at least one of them is defined; otherwise it is
+-- stuck.
+type family HasRep (s :: Type -> Type) (t :: Type -> Type) :: Bool
+type instance HasRep (s x) t = 'True
+type instance HasRep s (t x) = 'True
+
 -- | If no instance matches, GHC tends to bury error messages "No instance for
 -- LabelOptic..." within a ton of other error messages about ambiguous type
 -- variables and overlapping instances which are irrelevant and confusing. Use
--- overlappable instance providing a custom type error to cut its efforts short.
-instance {-# OVERLAPPABLE #-}
-  (LabelOptic name k s t a b,
-   TypeError
+-- incoherent instance providing a custom type error to cut its efforts short.
+instance {-# INCOHERENT #-}
+  TypeError
    ('Text "No instance for LabelOptic " ':<>: 'ShowType name
     ':<>: 'Text " " ':<>: QuoteType k
     ':<>: 'Text " " ':<>: QuoteType s
@@ -239,11 +269,8 @@ instance {-# OVERLAPPABLE #-}
     ':<>: 'Text " " ':<>: QuoteType a
     ':<>: 'Text " " ':<>: QuoteType b
     ':$$: 'Text "  (maybe you forgot to define it or misspelled a name?)")
-  ) => LabelOptic name k s t a b where
-  labelOptic = error "unreachable"
-
--- | Type synonym for a type-preserving optic as overloaded label.
-type LabelOptic' name k s a = LabelOptic name k s s a a
+   => GeneralLabelOptic name k s t a b has_rep where
+  generalLabelOptic = error "unreachable"
 
 instance
   (LabelOptic name k s t a b, is ~ NoIx
--- a/cabal.project
+++ b/cabal.project
@@ -7,6 +7,7 @@ packages:
   optics-vl/*.cabal
   indexed-profunctors/*.cabal
   template-haskell-optics/*.cabal
+  generic-optics-labels/*.cabal
 
 -- An example, using optics to generate optics diagrams
 packages:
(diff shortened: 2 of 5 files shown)


arybczak:
That is both brilliant and terrifying :smile: 

> It turns out that my worries about polymorphism are already an issue with `LabelOptic`: if you use a polymorphic expression such as `view #foo` without a type signature, you will hit the custom type error rather than remaining polymorphic in a `LabelOptic` constraint.

It already happens on `master`. It's not such a big deal there though since you need to explicitly write instances or use TH, but it becomes weird once generics come into play and valid instances can be "dynamically generated" for a lack of a better term.

I pushed a commit with a trick that prevents early instance resolution. It works as expected, though I feel we're creating a monster here, lol.

> If we go for this approach, I think we should rename `Optics.Label` to `Optics.Label.Core` in `optics-core` and make `optics` depend on `generic-optics-labels` so it can expose an `Optics.Label` module including the instance.

Sounds good :+1:

arybczak:
I ported support for generic prisms.

arybczak:
Note that if `optics` depends on `generic-optics-labels`, then it will require GHC 8.4.

optics-core/src/Optics/Internal/Optic.hs
```diff
@@ -210,25 +212,61 @@ class LabelOptic (name :: Symbol) k s t a b | name s -> k a
...
+-- below. Prevent that (if only to be able to inspect most polymorphic types of
+-- @#foo % #bar@ or @view #foo@ in GHCi), by defining a dummy instance that
+-- matches all names, thus postponing the reduction.
+instance LabelOptic name An_Iso Void0 Void0 Void0 Void0 where
+  labelOptic = Optic id
+
+-- | This type family should be called with applications of 'Rep' on both sides,
+-- and will reduce to 'True' if at least one of them is defined; otherwise it is
+-- stuck.
+type family HasRep (s :: Type -> Type) (t :: Type -> Type) :: Bool
```
arybczak: Perhaps it would be better for it to be called `RepDefined` and return `()` instead of Bool? considering that `False` is never used.

arybczak:
@adamgundry ping

> Note that if optics depends on generic-optics-labels, then it will require GHC 8.4.

`generic-optics` works fine with GHC 8.2.2 (even its tests pass), so the bounds can be relaxed. We talked about dropping support for 8.0.2, so I guess it's not that big of a deal.

I also implemented https://github.com/kcsongor/generic-lens/pull/114 so that generic lenses compile faster (and eat less memory).

adamgundry:
> I pushed a commit with a trick that prevents early instance resolution. It works as expected, though I feel we're creating a monster here, lol.

That's a nice trick, good idea. All this does feel a bit complex, but I can't currently see much wrong with it...

Thanks for adding prisms support. I wondered if that was possible but hadn't properly investigated. I'm lacking time at the moment but will try to properly grok how it works at some point soon.

I think bumping the minimum requirement for `optics` to GHC 8.2 (or even 8.4) is a reasonable trade-off at this point, especially if the pay-off is generic labels-as-optics (which is more compelling than #288 alone). Of course we could make the use of `generic-optics` conditional on having a new enough GHC, but I don't feel entirely comfortable with having the API depend on the GHC version in use.

At the risk of starting yet another bikeshedding round, should the adapter package be `generic-optics-labels` or just `optics-labels` (or even something like `optics-generic-labels`)? Something with `optics-` as a prefix makes it more obviously part of the `optics` stable.

optics-core/src/Optics/Internal/Optic.hs
adamgundry: This would work. I feel somewhat uncomfortable returning an informative `()` as morally we ought to have an eta-law for it (and perhaps GHC one day will, although if it does that would probably break a lot of the type family tricks). `RepDefined` is a better name than `HasRep` in any case.

arybczak:
> At the risk of starting yet another bikeshedding round, should the adapter package be `generic-optics-labels` or just `optics-labels` (or even something like `optics-generic-labels`)? Something with `optics-` as a prefix makes it more obviously part of the `optics` stable.

`optics-generic-labels` is a viable alternative. I don't mind current one though.

I consider `optics-labels` too general, it suggests it has THE way for labels, but sadly generic labels will (probably) always be inferior to hand-written/TH generated ones simply because they take longer to compile and generic representation is not optimized away for larger data types, so they are slower at both compile and run time.

With https://github.com/kcsongor/generic-lens/pull/114 and https://github.com/kcsongor/generic-lens/pull/116 generic lenses compile in a reasonable amount of time and both lenses and prisms are "only" twice as slow as TH generated ones when generic representation is not optimized away. That might be the best it gets, so they're still strictly inferior to TH generated ones.

arybczak:
After some pondering I'm for makng `optics-generic-labels` a part of `generic-optics` and not make `optics` depend on it.

I don't consider generic optics "production ready" for reasons I outlined above (even though I recently spent a lot of time experimenting and trying to make them work close to what TH gets us, even to the point of [patching GHC](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2965)) and I don't think we should include them in `optics`.

adamgundry:
@arybczak thanks for pushing on improving the performance of the generics stuff. I can see the argument that `optics` shouldn't put it front and centre, yet. On the other hand it would be nice to make it possible in some form. So should we aim to put `GeneralLabelOptic` in `optics-core-0.3` but avoid introducing the `optics -> generic-optics` dependency for now?

Then I guess the question is whether the generic `GeneralLabelOptic` instance should go in `generic-optics` (if @kcsongor is up for that) or if we should release an `optics-generic-labels`.

arybczak:
> On the other hand it would be nice to make it possible in some form. So should we aim to put `GeneralLabelOptic` in `optics-core-0.3` but avoid introducing the `optics -> generic-optics` dependency for now?

Yes, I think that's reasonable.

> Then I guess the question is whether the generic `GeneralLabelOptic` instance should go in `generic-optics` (if @kcsongor is up for that) or if we should release an `optics-generic-labels`.

Considering that `generic-lens` has labels within the package, including corresponding functionality in `generic-optics` should be alright. Unless @kcsongor wants to split it, but even then I'd argue `generic-optics-labels` belongs in the `generic-lens` repository ;)

arybczak:
Also, in principle right now we could do this:

```diff
diff --git a/optics-core/src/Optics/Internal/Optic.hs b/optics-core/src/Optics/Internal/Optic.hs
index 2e1418e..f9891e0 100644
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -41,10 +41,12 @@ import Data.Kind (Type)
 import Data.Proxy (Proxy (..))
 import Data.Type.Equality
 import GHC.OverloadedLabels
+import GHC.Records
 import GHC.TypeLits
 
 import Data.Profunctor.Indexed
 
+import Optics.Internal.Bi
 import Optics.Internal.Optic.Subtyping
 import Optics.Internal.Optic.TypeLevel
 import Optics.Internal.Optic.Types
@@ -225,22 +227,16 @@ instance
   ) => LabelOptic name k Void0 Void0 a b where
   labelOptic = Optic id
 
--- | If no instance matches, GHC tends to bury error messages "No instance for
--- LabelOptic..." within a ton of other error messages about ambiguous type
--- variables and overlapping instances which are irrelevant and confusing. Use
--- overlappable instance providing a custom type error to cut its efforts short.
+-- | If no instance matches, leverage 'HasField' machinery to generate a
+-- 'Optics.Getter.Getter' for field access.
 instance {-# OVERLAPPABLE #-}
-  (LabelOptic name k s t a b,
-   TypeError
-   ('Text "No instance for LabelOptic " ':<>: 'ShowType name
-    ':<>: 'Text " " ':<>: QuoteType k
-    ':<>: 'Text " " ':<>: QuoteType s
-    ':<>: 'Text " " ':<>: QuoteType t
-    ':<>: 'Text " " ':<>: QuoteType a
-    ':<>: 'Text " " ':<>: QuoteType b
-    ':$$: 'Text "  (maybe you forgot to define it or misspelled a name?)")
+  ( HasField name s a
+  , k ~ A_Getter
+  , s ~ t
+  , a ~ b
   ) => LabelOptic name k s t a b where
-  labelOptic = error "unreachable"
+  labelOptic = Optic (lmap (getField @name @s @a) . rphantom)
+  {-# INLINE labelOptic #-}
 
 -- | Type synonym for a type-preserving optic as overloaded label.
 type LabelOptic' name k s a = LabelOptic name k s s a a
```

to get field getters for free (and lenses once `HasField` gets a setter). However, if we add `generalLabelOptic`, we can't have `optics` pull in the instance for serving generic optics as then for data types having an instance of `Generic` the above `LabelOptic` instance becomes unreachable :thinking:

arybczak:
I made a PR for generic-lens that adds support for labels: https://github.com/kcsongor/generic-lens/pull/120

I believe this PR can be now closed.

PR #291, opened 2020-03-07
Make optics break less with simplified subsumption

phadej:
Patch mostly by Ryan Scott, see
https://gitlab.haskell.org/RyanGlScott/head.hackage/blob/9fadfda472c80007e7a52471cf891587aa6b7465/patches/optics-core-0.2.patch

I rewrote `%%` def to make type-checker guess less

### DIFF
--- a/codegen/Subtypes.hs
+++ b/codegen/Subtypes.hs
@@ -146,7 +146,7 @@ genSubtypes = either (fail . show) id $ runG opticsKind $ \g -> do
                     [ "instance Is"
                     , leftpad (show k)
                     , leftpad (show k')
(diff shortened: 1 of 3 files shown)


arybczak:
What does it help with?

phadej:
https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0287-simplify-subsumption.rst will break `optics-core` (and `optics-extra` too, but I omitted that for now).

The changes in this PR should make `optics-core` work even when simplified subsumption is merged to GHC. I don't see any drawbacks either, as CI is green.

PR #294, opened 2020-03-11
Add filteredBy/unsafeFilteredBy

### LINKED ISSUE #293: Add filteredBy / unsafeFilteredBy
Would be nice to add filteredBy as seen in `lens`. I am using this definition internally, hope it's correct:
```haskell
-- Use an Affine fold as a filter
filteredBy :: Is k An_AffineFold  => Optic' k NoIx a i -> IxFold i a a
filteredBy p = ifoldVL $ \f val -> case preview p val of
  Nothing -> pure val
  Just witness -> f witness val

-- Use an Affine fold as a filter. Not always a traversal, refer filtered
unsafeFilteredBy :: Is k An_AffineFold  => Optic' k NoIx a i -> IxTraversal' i a a
unsafeFilteredBy p = itraversalVL $ \f val -> case preview p val of
  Nothing -> pure val
  Just witness -> f witness val
```

Bonus: I've found this optic useful in generating indexes from a field in a record:
```haskell
-- Use an optic as an index on our value
itraverseBy :: Is k A_Getter => Optic' k NoIx s i-> IxTraversal' i s s
itraverseBy idx = itraversalVL $ \f s -> f (view idx s) s
```

arybczak:
Fixes #293.

Note that the first commit breaks API as it renames two existing functions, but they're obscure enough that I'm pretty sure (almost) no one has used them.

Their new names correspond nicely with appropriate Fold/Traversal functions, so IMO it's worth it.

### DIFF
--- a/optics-core/src/Optics/AffineFold.hs
+++ b/optics-core/src/Optics/AffineFold.hs
@@ -26,9 +26,11 @@ module Optics.AffineFold
   -- @
 
   -- * Additional introduction forms
+  , afoldVL
   , filtered
 
   -- * Additional elimination forms
+  , atraverseOf_
   , isn't
 
     -- * Semigroup structure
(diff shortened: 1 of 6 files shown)


phadej:
Breaking changes are fine. We already have some since `0.2` (e.g. `mapping`, `element` type changes).

PR #300, opened 2020-03-27
Document monoidal structures of Folds

### LINKED ISSUE #244: Document monoidal structure of folds

adamgundry:
Fixes #244. I came across some `lens` code using `<>` and wondered what the equivalent in `optics` would look like...

### DIFF
--- a/optics-core/src/Optics/AffineFold.hs
+++ b/optics-core/src/Optics/AffineFold.hs
@@ -33,7 +33,16 @@ module Optics.AffineFold
   , atraverseOf_
   , isn't
 
-    -- * Semigroup structure
+  -- * Monoid structure
+  -- | 'AffineFold' admits a monoid structure where 'afailing' combines folds
+  -- (returning a result from the second fold only if the first returns none)
+  -- and the identity element is 'Optics.IxAffineTraversal.ignored' (which
+  -- returns no results).
+  --
+  -- /Note:/ There is no 'Optics.Fold.summing' equivalent that returns an
+  -- 'AffineFold', because it would not need to return more than one result.
+  --
+  -- There is no 'Semigroup' or 'Monoid' instance for 'AffineFold'.
   , afailing
 
   -- * Subtyping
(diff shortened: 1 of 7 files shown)


arybczak:
Looks good, thanks :+1: Could you also do this for `(Ix)AffineFold`?

adamgundry:
Sure, good idea.

Although hang on... I wrote that `failing` does not form a monoid on folds, but actually isn't `ignored` an identity for `failing` as well as `summing`? After all:
```hs
ignored `failing` x == x -- from definition of failing, as ignored returns no results

-- case split on whether x returns any results
x `failing` ignored == x -- if x returns at least one result, from definition of failing
x `failing` ignored == ignored -- if x returns no results, but then x == ignored
```
So aren't they both monoids? Or am I missing something?

arybczak:
Looks like that's the case, I missed that :thinking:

arybczak:
Makes sense, `summing` vs `failing` is like `[]` vs `Maybe`, and both are monoids where neutral element is `[]`/`Nothing`, so in this case neutral element is an optic that returns no results :)

adamgundry:
I've updated the docs to talk about monoids instead of semigroups throughout, and made one or two other minor tweaks.

optics-core/src/Optics/AffineFold.hs
```diff
@@ -33,7 +33,16 @@ module Optics.AffineFold
...
+  -- * Monoid structure
+  -- | 'AffineFold' admits a monoid structure where 'afailing' combines folds
+  -- (returning a result from the second fold only if the first returns none)
+  -- and the identity element is 'Optics.IxAffineTraversal.ignored' (which
+  -- returns no results).
+  --
+  -- /Note:/ There is no 'Optics.Fold.summing' equivalent that returns an
+  -- 'AffineFold', because it would not need to return more than one result.
+  --
+  -- There is no 'Semigroup' or 'Monoid' instance for 'AffineFold'.
```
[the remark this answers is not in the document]
adamgundry: Good idea, done.

optics-core/src/Optics/Fold.hs
```diff
@@ -76,7 +76,7 @@ module Optics.Fold
   , backwards_
 
   -- * Monoid structures
-  -- | 'Fold' admits (at least) two monoid structures:
+  -- | 'Fold' admits (at least) two monoid structures: #monoids#
```
phadej, line 79: what's that `#monoids#` is?

adamgundry: Haddock syntax for an anchor, so I can link to this section from elsewhere via `"Optics.Fold#monoids"`.

phadej: TIL

PR #301, opened 2020-03-27
Generalise ignored to be an IxAffineTraversal instead of an IxTraversal

adamgundry:
Optics.IxTraversal still exports ignored, for backwards compatibility.


phadej:
Force pushed an amendment to fix doctests

adamgundry:
Thanks!

PR #302, opened 2020-03-30
Add optics-sop and metametapost GHC-8.10 travis jobs

phadej:


.travis.yml
```diff
@@ -116,12 +116,30 @@ install:
...
-    if $GHCJS || ! $GHCJS && [ $HCNUMVER -lt 81000 ] ; then echo "packages: optics-sop" >> cabal.project ; fi
+    echo "packages: optics-sop" >> cabal.project
     echo "packages: optics-th" >> cabal.project
     echo "packages: optics-vl" >> cabal.project
     echo "packages: indexed-profunctors" >> cabal.project
     echo "packages: template-haskell-optics" >> cabal.project
-    if ! $GHCJS && [ $HCNUMVER -lt 81000 ] ; then echo "packages: metametapost" >> cabal.project ; fi
+    if ! $GHCJS ; then echo "packages: metametapost" >> cabal.project ; fi
+  - echo 'package optics' >> cabal.project
+  - "echo '  ghc-options: -Werror=missing-methods' >> cabal.project"
```
phadej: These would catch e.g. missing `liftTyped` in `Lift` class, have we any such instances.

arybczak: Why not add it to ghc-options in `.cabal` files?

phadej: Because `haskell-ci` (re)generates it for me. Each `.cabal` file needs manual editing.

I'm not fan of adding `ghc-options: -Wall` to `.cabal` files either. It's irrelevant for library users. As soon as there is concise way to configure `ghc-options` for all local packages, I'd move them to `cabal.project`.

PR #303, opened 2020-03-31
Get rid of warnings (including GHC 8.11.0)

arybczak:


phadej:
I pushed a commit. Code looks nicer that way.

PR #304, opened 2020-04-01
Add GeneralLabelOptic for pluggable generic optics as labels

arybczak:
Extracts from #290 stuff needed to plug in generic optics as labels.

I modified it slightly, i.e. added `RepDefined` type as a witness for clarity.

Looks like https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2965 will go in after a few adjustments, so generic optics might be a viable choice (eventually).

### DIFF
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -30,6 +30,7 @@ module Optics.Internal.Optic
   -- * Labels
   , LabelOptic(..)
   , LabelOptic'
+  , GeneralLabelOptic(..)
   -- * Re-exports
   , module Optics.Internal.Optic.Subtyping
   , module Optics.Internal.Optic.Types
@@ -40,6 +41,7 @@ import Data.Function ((&))
 import Data.Kind (Type)
 import Data.Proxy (Proxy (..))
 import Data.Type.Equality
+import GHC.Generics (Rep)
 import GHC.OverloadedLabels
 import GHC.TypeLits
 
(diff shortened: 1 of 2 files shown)


adamgundry:
This looks reasonable to me for the present. I wonder if we should advertise in the docs somewhere that this might be subject to change in the future? I can see two plausible futures, which might happen in any order or not at all:
 * Generics become good enough (or we change our minds) that we can adopt generic optics more centrally (i.e. in `optics`). This probably wouldn't change `GeneralLabelOptic` assuming we wouldn't include generic optics in `optics-core`.
 * `HasField` gains (type-changing?) setters, in which case we might just want to use it directly as the catch-all `LabelOptic` instance and drop `GeneralLabelOptic`.

arybczak:
>     * `HasField` gains (type-changing?) setters, in which case we might just want to use it directly as the catch-all `LabelOptic` instance and drop `GeneralLabelOptic`.

This doesn't necessarily has to happen if generic-optics also switches lens instances to use `HasField` from `GHC.Records`, then both paths (when a data type has Generic instance and when it doesn't) will have the same result.

> I wonder if we should advertise in the docs somewhere that this might be subject to change in the future?

We could, but as far as I'm aware there's an implicit assumption (we might want to write it down somewhere though) that anything in internal modules that is not reexported in public ones is subject to change :thinking:

phadej:
@arybczak there's conflict now.

phadej:
Conflict resolved in https://github.com/well-typed/optics/commit/7a085f861c5667099ffe71a7c44e85d7a874453c

Issue #309, opened 2020-04-10
Split {Functor,Foldable,Traversable}WithIndex to a separate package


arybczak:
These can be made a standalone package easily.

Then:
- `unordered-containers` and `vector` can supply instances instead of `optics-extra`
- people can use it without pulling in optics as the classes are useful on their own
- `lens` can switch to using these classes so that we don't duplicate them (I saw claims that they need to have `ifolded`, `itraversed` and `imapped` as class member for performance, but I doubt that's the case considering that we don't do that and have the same performance in benchmarks as `lens`).

Thoughts?

phadej:
I'd like to have that package, especially if we get `lens` use it. Otherwise I think it would be useless.

If `lens` would also use `map-with-index` (because `indexed-functors` is something different, given `indexed-profunctors`), then there might be enough "pressure" to reverse the dependency so `unordered-containers` and `vector` would depend on it, not the other way around. This is justified as it is more likely (though highly unlikely) that `FunctorWithIndex` would be in `base` than that the `HashMap` would be there.

So, I think we have to convince Edward (about the performance) first before making the package. I also think that `ifolded` in `lens` don't need to be a member, but it's just a hunch.

adamgundry:
I think this might be useful to packages other than `lens`, and if someone wants it then splitting it out makes sense, but I agree the motivation is much more compelling if `lens` itself can use it. (I wonder if any other lens libraries might benefit from it? My first thought was `microlens` but that doesn't seem to have indexed optics.)

ocharles:
Is this perhaps a resurrection of Edward's https://hackage.haskell.org/package/keys-3.12.3 library?

arybczak:
Right, it looks like it would be a subset of it. Most of the dependencies of this package would have to be reversed though.

phadej:
This is done now.

PR #310, opened 2020-04-10
Move indexed classes to a separate module, add itoList

arybczak:
Good for separation of concerns and potential prelude for #309.


optics-core/optics-core.cabal
```diff
@@ -97,7 +97,8 @@ library
                    -- internal modules
                    Optics.Internal.Bi
                    Optics.Internal.Fold
-                   Optics.Internal.Indexed
+                   Optics.Internal.Indexed.Classes
+                   Optics.Internal.Indexed.Utils
```
phadej, line 101: I'd keep `Utils` as `Optics.Internal.Indexed`, given the `Optics.Internal.Indexed.Classes` could be some `Data.Traversable.WithIndex` ... later.

arybczak: Fixed :+1:

optics-core/src/Optics/Internal/Indexed/Classes.hs
```diff
@@ -0,0 +1,514 @@
...
+  ifoldMap f = getConst #. itraverse (\i a -> Const (f i a))
+  {-# INLINE ifoldMap #-}
+
+  ifoldr :: (i -> a -> b -> b) -> b -> f a -> b
+  ifoldr iabb b0 = (\e -> appEndo e b0) . ifoldMap (\i -> Endo #. iabb i)
+  {-# INLINE ifoldr #-}
+
+  ifoldl' :: (i -> b -> a -> b) -> b -> f a -> b
+  ifoldl' ibab b0 s = ifoldr (\i a bb b -> bb $! ibab i b a) id s b0
+  {-# INLINE ifoldl' #-}
```
phadej, line 58: I wonder if `itoList` should be the member class.

```haskell
instance FoldableWithIndex k (Map.Map k v) where
    itoList = Map.toList
```

Not an issue for this PR, more of a note for #309

arybczak: Makes sense, I presume we'd want to mimic `Foldable` with the amount of methods if it's to be split.

phadej: Yes, it might make sense to add all four of `ifoldr`, `ifoldr'`, `ifoldl'`, and `ifoldl`. And then have `ifoldMap` and `itoList`, so six in total.

PR #312, opened 2020-04-11
Release 0.3

adamgundry:
I propose ~~Monday~~Wednesday as the release date and have updated things appropriately and uploaded tentative package candidates. ~~The only significant change is that I propose to release `optics-vl` as `0.3` instead of `0.2.1`. While in theory the latter is permissible, it doesn't seem worth deviating from the version numbering of the other packages at this point. Please shout if you disagree.~~

* https://hackage.haskell.org/package/optics-core-0.3/candidate
* https://hackage.haskell.org/package/optics-extra-0.3/candidate
* https://hackage.haskell.org/package/optics-th-0.3/candidate
* https://hackage.haskell.org/package/optics-0.3/candidate
* https://hackage.haskell.org/package/optics-vl-0.2.1/candidate

### DIFF
--- a/optics-core/CHANGELOG.md
+++ b/optics-core/CHANGELOG.md
@@ -1,19 +1,25 @@
-# optics-core-0.3 (TBD)
+# optics-core-0.3 (2020-04-15)
 * GHC-8.10 support
 * Add `filteredBy` and `unsafeFilteredBy`
-* Add indexed instances for `Const` and `Constant`
+* Add `FunctorWithIndex`, `FoldableWithIndex` and `TraversableWithIndex`
+  instances for `Const` and `Constant`
 * Add `afoldVL` and `iafoldVL` constructors
 * Rename `toAtraversalVL` to `atraverseOf`, and `toIxAtraversalVL` to `iatraverseOf`
-* `element` and `elementOf` construct `(Ix)AffineTraversal`
-* Change `mapping` to work also on `Lens` and `Prism` degenerating to `Getter` and `Review`.
+* Generalise `element` and `elementOf` to construct `IxAffineTraversal`s
+  instead of `IxTraversal`s
+* Change `mapping` to work on optic kinds other than `Iso`: it now supports
+  `Lens` and `Prism` degenerating to `Getter` and `Review` respectively
 * Generalise `ignored` to be an `IxAffineTraversal` instead of an `IxTraversal`
 * Add `singular` and `isingular`
 * Add `(^?!)` operator
 * Expose `Curry` and `CurryCompose`
 * Show expected elimination forms on optic kind mismatch
-* Add GeneralLabelOptic for pluggable generic optics as labels
-* Document monoidal structures of Folds
-* Remove proxy argument from implies
+* Use stricter `uncurry'` for better performance
+* Add hidden `LabelOptic` instance to postpone instance resolution
+* Add `GeneralLabelOptic` for pluggable generic optics as labels
+* Document monoidal structures of `Fold`s
+* Remove proxy argument from `implies`
+* Add `itoList`
 
 # optics-core-0.2 (2019-10-18)
 * Add `non`, `non'` and `anon` to `Optics.Iso`
--- a/optics-core/src/Optics/IxAffineFold.hs
+++ b/optics-core/src/Optics/IxAffineFold.hs
@@ -118,6 +118,8 @@ iafolding g = iafoldVL (\point f s -> maybe (point s) (uncurry' f) $ g s)
 
 -- | Obtain a potentially empty 'IxAffineFold' by taking the element from
 -- another 'AffineFold' and using it as an index.
+--
+-- @since 0.3
 filteredBy :: Is k An_AffineFold  => Optic' k is a i -> IxAffineFold i a a
 filteredBy p = iafoldVL $ \point f s -> case preview p s of
   Just i  -> f i s
--- a/optics-extra/CHANGELOG.md
+++ b/optics-extra/CHANGELOG.md
@@ -1,12 +1,10 @@
-# optics-extra-0.3 (TBD)
+# optics-extra-0.3 (2020-04-15)
 * `optics-core-0.3` compatible release
 * GHC-8.10 support
-* Use stricter uncurry for better performance
+* Use stricter `uncurry'` for better performance
 
-# optics-extra-0.2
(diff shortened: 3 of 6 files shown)


arybczak:
One question: why do `optics-extra` and `optics` have upper bounds for optics-family packages set to `<0.3.1`, but `optics-th` and `optics-vl` has them set to `0.4`?

Wouldn't it be better to set all of them to `<0.4`?

phadej:
`optics-extra` could.

`optics` re-exports `optics-core`, `optics-extra`, and `optics-th`functionality so it must have tight bounds on them. This way we ensure that `optics` interface doesn't change without that `optics` version also changes. For example users don't need to *also* specify `optics-core >=0.3.1` bound to ensure that some new combinator is present which was added in `optics-core-0.3.1`, but is available through `optics` re-exports.

phadej:
> optics-vl as 0.3 instead of 0.2.1.

I don't have strong opinion about that yet, but OTOH I do think that `optics-vl` is close to "done", so would rather let is have whatever version it has now.

arybczak:
What about bugfix releases? Would it be 0.3.0.1?

phadej:
The `build-depends: optics-core >=0.3 && <0.3.1` in `optics-0.3` would accept `optics-core-0.3.0.1` if such would be done. Though in practice such releases are often ones which have some inner changes to e.g. accommodate new version of a dependency (which `optics-core` doesn't really have).

If we would make any changes to `.Internal` changes, those **must** be minor releases. Then we'd need to make a revision for `optics` allowing new `optics-core` minor version too. That's the price for exposing internals in the first place. One reason is that CPP allows to case on three first version digits, also in practice if someone uses `.Internal` modules, it would be **required** to specify exact versions as `==0.3`, which is impractical. `optics-core <0.3.1` is a compromise - a same requirement as if you define orphan instances, e.g.

Personally I would rather consider `Internal` modules as a part of "stable" API, and simply don't change them - just advice to avoid them if one could. Because I don't think that other packages which use internal modules would use stricter upper bounds in their dependency definitions, so if we actually change something, we'll break them. But that's another issue.

adamgundry:
> If we do this, then the lower-bound on `optics-core` dependency have to be `optics-core >=0.3` too. Morally.

Why? 

> `optics-0.2.1` is IMHO better because if someone is "stuck" with `optics-core-0.2` they could use `optics-vl-0.2.1`.
> 
> Or to put it differently. We could release the current status of `optics-vl` with `optics-core <0.3` bound still, without releasing `optics-0.3` - it's standalone. And later when `optics` is released, the we can make revision to allow `optics-0.3`
> 
> TL;DR it would be better if `optics-vl` and `optics-core` / `optics-extra` / `optics` version numbers (and release cycles) are decoupled. The more they drift from each other the less confusion there will be.
>
> EDIT: it might be (in theory, more applicable to `optics-sop`) that one would be forced to make major release of `optics-vl` or `optics-sop` due changes in non-`optics` (i.e. `generics-sop-0.6`)

Okay, you've convinced me. I was thinking it would lead to much confusion if we had inconsistent numbering between e.g. `optics-core` / `optics-extra` / `optics`, but that doesn't necessarily apply to `optics-vl` or other packages. So I'll revert to releasing `optics-vl-0.2.1`. 

>> One question: why do optics-extra and optics have upper bounds for optics-family packages set to <0.3.1, but optics-th and optics-vl have them set to <0.4?
>>
>> Wouldn't it be better to set all of them to <0.4?
> 
> `optics-extra` could.

I don't think so, because `Optics.Indexed` re-exports modules from `optics-core`, so its API depends on the `optics-core` minor version.

How does this sound as a summary/policy for future releases?
 * `optics-core`, `optics-extra` and `optics` are all "coupled", i.e. they are always released together with the same minor version, and have bounds that require the same minor version. They may have bugfix releases independently, but only if there are essentially no API changes. (Perhaps one day they will all be one package using multiple public libraries.)
 * `optics-th` can have looser bounds on `optics-core`, but `optics` must specify a minor version bound on `optics-th`. (Should it have its own version numbering/release schedule, or should it match `optics`?)
 * `optics-vl` is released independently on its own schedule, and usual PVP rules apply.

phadej:
> I don't think so, because `Optics.Indexed`...

Good point. I missed that.

Your summary/policy is good. About `optics-th`: `Optics.TH` is re-exported in `optics`, so I'd say all four (`optics-core`, `optics-extra`, `optics-th` and `optics`) are "coupled".

phadej:
It's somewhat expected that `optics-th` would change mostly due `template-haskell` changes (hopefully). Then it's IMHO ok to have

If we start with all four being `0.3`, there is an addition in `optics-core` which propagates:

```
optics-core-0.3.1
optics-extra-0.3.1
optics-th-0.3
optics-0.3.1
```

and then if change occurs in ` optics-th` we'll do:

```
optics-core-0.3.1
optics-extra-0.3.1
optics-th-0.3.2
optics-0.3.2
```

Which is, all four packages would share the major series, and "minor version counter", but we could avoid re-uploading essentially the same package with another version.

I think that would be useful, even requires to be a bit more careful, because libraries I maintain depend on `optics-core` and `optics-extra` only, so avoiding new version numbers will keep build caches warm.

I must admit, I'm annoyed when e.g. new release of `distributive` causes cascade rebuilding `lens`, or `dlist` busting `aeson` caches. Sometimes the releases are unavoidable, sometimes they are.

OTOH, those are theoretical concerns. `optics` is not moving with even release every month pace, and it is unluckily it would, so above can be simplified if the release pace constraint is true.

adamgundry:
I think this is basically ready, though I need to re-upload `optics-vl`. But @arybczak would you prefer to hang on a little longer to land your `optics-th` changes?

arybczak:
> @arybczak would you prefer to hang on a little longer to land your optics-th changes?

That would be lovely. It should be ready tomorrow.

phadej:
Let us have a timeout with #313 and #314. I don't like "let's get them in quickly to get them into the release". That's famous last words.

I'm unhappy that 0.3 release is delayed, but let's do stuff properly.

arybczak:
I propose we include #314 and postpone #313 until next release.

#314 is trivial, #313 not so much.

arybczak:
Now I think that #313 is alright to go in as it fixes a few bugs.

adamgundry:
I've boldly merged #313, updated the changelogs and will release soon unless anyone objects.

PR #313, opened 2020-04-13
Make generation of LabelOptic instances aware of injective type families  and polymorphic kinds

arybczak:
Existing check wasn't taking injective type families into account (and wasn't working properly really).

### DIFF
--- a/optics-th/optics-th.cabal
+++ b/optics-th/optics-th.cabal
@@ -54,6 +54,7 @@ test-suite optics-th-tests
   build-depends: base
                , optics-core
                , optics-th
+               , tagged
 
   type:    exitcode-stdio-1.0
   main-is: Optics/TH/Tests.hs
--- a/optics-th/src/Language/Haskell/TH/Optics/Internal.hs
+++ b/optics-th/src/Language/Haskell/TH/Optics/Internal.hs
@@ -1,10 +1,13 @@
-{-# LANGUAGE LambdaCase       #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE LambdaCase #-}
 module Language.Haskell.TH.Optics.Internal
   (
   -- * Traversals
     HasTypeVars(..)
   , typeVars      -- :: HasTypeVars t => Traversal' t Name
+  , typeVarBndrs
   , substTypeVars -- :: HasTypeVars t => Map Name Name -> t -> t
+  , SubstType(..)
 
   -- * Prisms
   , _FamilyI
@@ -15,6 +18,7 @@ module Language.Haskell.TH.Optics.Internal
 
 import Data.Map as Map hiding (map, toList)
 import Data.Maybe (fromMaybe)
+import Data.Foldable (traverse_)
 import Data.Set as Set hiding (map, toList)
 import Language.Haskell.TH
 
@@ -67,6 +71,11 @@ instance HasTypeVars Type where
                                  <*> pure n
                                  <*> traverseOf (typeVarsEx s) f t2
     ParensT t         -> ParensT <$> traverseOf (typeVarsEx s) f t
+#if MIN_VERSION_template_haskell(2,15,0)
+    AppKindT t k       -> AppKindT <$> traverseOf (typeVarsEx s) f t
+                                   <*> traverseOf (typeVarsEx s) f k
+    ImplicitParamT n t -> ImplicitParamT n <$> traverseOf (typeVarsEx s) f t
+#endif
     t                 -> pure t
 
 instance HasTypeVars t => HasTypeVars [t] where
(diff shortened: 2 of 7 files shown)


arybczak:
@adamgundry looks like with type-changing updates we also have to worry about changing kinds, though it seemed easy enough to handle here :thinking:

phadej:
@arybczak isn't it that when you write those by hand, the kind variables are usually implicitly quantified so one doesn't really worry?

Related to kinds, @arybczak have you noted https://github.com/ekmett/lens/issues/917, does `optics-th` has the same problem?

arybczak:
> @arybczak isn't it that when you write those by hand, the kind variables are usually implicitly quantified so one doesn't really worry?

I'm not sure what you mean.

> Related to kinds, @arybczak have you noticed [ekmett/lens#917](https://github.com/ekmett/lens/issues/917), does `optics-th` have the same problem?

Yes, mostly. The example from the first post works when you replace `makePrisms` with `makePrismLabels`, but the rest fails.

phadej:
> > @arybczak isn't it that when you write those by hand, the kind variables are usually implicitly quantified so one doesn't really worry?
> 
> I'm not sure what you mean.

```haskell
{-# LANGUAGE DataKinds             #-}
{-# LANGUAGE FlexibleInstances     #-}
{-# LANGUAGE GADTs                 #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE OverloadedLabels      #-}
{-# LANGUAGE PolyKinds             #-}
{-# LANGUAGE UndecidableInstances  #-}

import           Data.Proxy
import           GHC.TypeLits
import           Optics.Core

instance (k ~ An_Iso, u ~ Proxy a, v ~ Proxy b) => LabelOptic "proxy" k (Proxy a) (Proxy b) u v where
    labelOptic = iso (const Proxy) (const Proxy)


proxy5 :: Proxy 5
proxy5 = Proxy

proxyFoo :: Proxy "foo"
proxyFoo = proxy5 & #proxy .~ Proxy
```

Here the instance is polykinded, isn't it?

If I write

```haskell
instance (k ~ An_Iso, u ~ Proxy a, v ~ Proxy b) => LabelOptic "proxy" k (Proxy (a :: ak)) (Proxy (b :: ak)) u v where
```

an example doesn't compile. 

If I write

```haskell
instance (k ~ An_Iso, u ~ Proxy a, v ~ Proxy b) => LabelOptic "proxy" k (Proxy (a :: ak)) (Proxy (b :: bk)) u v where
```

Then it does again.

I actually don't know which would be better, first or last (implicit or explicit). The same question as in lens 917.

arybczak:
Right. I still don't know how that relates to what's in the PR :) You mean type signatures in tests? I've written them explicitly with changing kinds to make it easier to see what's going on.

phadej:
It doesn't relate to this PR, it was more a comment along lines: "if we can change types, then changing kinds shouldn't be much more difficult - if at all"

adamgundry:
This looks like a step in the right direction to me.

Thinking about it, I came up with this example:
```hs
data FamRec8 k a = FamRec8
  { _famRec8Thing :: Proxy (a :: k)
  }
makeFieldLabels ''FamRec8
```
which generates
```hs
instance forall a1 k (a2 :: k) b (a3 :: k).
         (a1 ~ Proxy a2, b ~ Proxy a3) =>
         LabelOptic "thing" An_Iso (FamRec8 k a2) (FamRec8 k a3) a1 b
```
but this doesn't permit kind-changing, when it could, because the manually written instance is accepted:
```hs
instance (u ~ Proxy a, v ~ Proxy b) => LabelOptic "thing" An_Iso (FamRec8 k a) (FamRec8 l b) u v where
  labelOptic = iso _famRec8Thing FamRec8
```

arybczak:
Interesting. Looks like TH doesn't give enough information to do that.

Within TH
```haskell
data FamRec8 k a = FamRec8
  { _famRec8Thing :: Proxy (a :: k)
  }
```
and
```haskell
data FamRec8 k a = FamRec8
  { _famRec8Thing :: Proxy a
  }
```
are indistinguishable :confused: And in the second case `k` is a phantom type.

phadej:
What TH tells about:

```haskell
data FamRec8 k (a :: k) = FamRec8
  { _famRec8Thing :: Proxy a
  }
```

arybczak:
Never mind, it required a bit of gymnastics, but is supported now.

I also made the first example from ekmett/lens#917 compile. Other ones are a bit more tricky, but also can be fixed.

arybczak:
Now all examples from ekmett/lens#917 work :+1:

arybczak:
Ok, it's ready to be merged. Changes:
1) TH generated `LabelOptic` instances for field optics work properly in presence of type families
2) Improved support for TH generated optics in presence of kind polymorphism
3) Fixed calculation of phantom types in TH generation of `LabelOptic` prism instances

PR #314, opened 2020-04-13
Make optic kind a type equality for better type inference / errors

arybczak:
```haskell
data Test = Test { testX :: Int }
makeFieldLabels ''Test

x :: Lens' Test Int
x = #x
```

Before:
```
    • No instance for LabelOptic "x" ‘A_Lens’ ‘Test’ ‘Test’ ‘Int’ ‘Int’
        (maybe you forgot to define it or misspelled a name?)
    • In the expression: #x
      In an equation for ‘x’: x = #x
```

After:
```
    • Couldn't match type ‘A_Lens’ with ‘An_Iso’
        arising from the overloaded label ‘#x’
    • In the expression: #x
      In an equation for ‘x’: x = #x
```

### DIFF
--- a/optics-th/src/Optics/TH/Internal/Product.hs
+++ b/optics-th/src/Optics/TH/Internal/Product.hs
@@ -145,21 +145,23 @@ makeFieldLabel
 makeFieldLabel rules (defName, (defType, cons)) = do
   (context, instHead) <- case defType of
     OpticSa _ _ otype s a -> do
+      (k,  cxtK) <- eqSubst (ConT $ opticTypeToTag otype) "k"
       (a', cxtA) <- eqSubst a "a"
       (b', cxtB) <- eqSubst a "b"
-      pure (pure [cxtA, cxtB], pure $ conAppsT ''LabelOptic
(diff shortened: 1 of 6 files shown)


optics-th/src/Optics/TH/Internal/Product.hs
```diff
@@ -145,21 +145,23 @@ makeFieldLabel
 makeFieldLabel rules (defName, (defType, cons)) = do
   (context, instHead) <- case defType of
     OpticSa _ _ otype s a -> do
+      (k,  cxtK) <- eqSubst (ConT $ opticTypeToTag otype) "k"
```
phadej, line 148: Could you write a description what kind of instance this code generates. I *guess* you make `k ~ A_Lens` stuff, but I don't like guessing...

Or you could take the above as a hint to update `Optics.Label` documentation. `metametapost` also have two instances written out by hand, which could be updated to show "best practices".

arybczak:
Updated.

phadej:
Ping @adamgundry that's something worth a changelog entry, probably, even shouldn't in practice affect anyone.

PR #317, opened 2020-05-05
Handle nullary type families properly

arybczak:


phadej:
Could you add a test case?

arybczak:
It's already there.

PR #325, opened 2020-06-21
Optimize away profunctor classes when profiling (#324)

### LINKED ISSUE #324: GHC doesn't optimize away profunctor type classes with profiling enabled
Hi,

I tried optics by switching over a project from lens. This worked very well due to the good type errors. I also think `%` is not bad and actually helps readability, since it is made clear where an optic is constructed. After finishing that I did some profiling using `+RTS -p`. Now `Profunctor.linear` and other optics-related functions appear prominently at the top of the profile. Is this some profiling artifact or is there some inlining issue present? Is the profunctor representation less efficient than the VL representation?

Another unrelated question - Many operators like `+~`, `<>~` seem to be missing from optics? Is this intentional since those can be replicated trivially or due you generally encourange a less operator-heavy style, where not every operator is replicated for lens modifications? 

Please let me know if you have better place to ask such questions! Thank you!

arybczak:
~~Fixes #324.~~

Almost.

### DIFF
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -22,6 +22,7 @@ module Optics.Internal.Optic
   , Optic'
   , Optic_
   , Optic__
+  , getOptic
   , castOptic
   , (%)
   , (%%)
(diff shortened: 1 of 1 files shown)


optics-core/src/Optics/Internal/Optic.hs
```diff
@@ -66,10 +67,19 @@ import Unsafe.Coerce (unsafeCoerce)
...
+-- | Strip the newtype wrapper off.
+getOptic
+  :: Profunctor p
+  => Optic k is s t a b
+  -> Optic_ k p i (Curry is i) s t a b
+-- Note: This is not part of the definition of 'Optic' because it needs to be
+-- marked INLINE for GHC to optimize away profunctor classes when profiling.
+-- See https://github.com/well-typed/optics/issues/324 for more details.
+getOptic (Optic o) = o
+{-# INLINE getOptic #-}
```
phadej, line 82: doesn't `{-# INLINE getOptic #-}` just work when `getOptic` (selector) function is defined with record syntax?

minad:
The newtype unwrap function is the issue? 👀 I will check if this works for me!

optics-core/src/Optics/Internal/Optic.hs
minad: I think the inline should not be necessary, since newtypes are advertised as "zero cost" wrappers. So one would expect those to be automatically inline. Afaik the newtypes are compiled via coercions. If this gets in the way of optimizations, this feels even more like something GHC could improve.

arybczak: > doesn't {-# INLINE getOptic #-} just work when getOptic (selector) function is defined with record syntax? 

It doesn't seem possible to do that then.

> I think the inline should not be necessary, since newtypes are advertised as "zero cost" wrappers.

Yes, it's incredibly bizarre. However, removing the INLINE pragma doesn't improve the situation.

I think it might be connected to this documentation bit:

> -fprof-auto
>
> All bindings not marked INLINE, whether exported or not, top level or nested, will be given automatic SCC annotations. Functions marked INLINE must be given a cost centre manually.

arybczak:
I'll merge after the weekend unless anyone objects.

PR #326, opened 2020-07-07
Improve documentation and add guide on how to use labels as optics

arybczak:


arybczak:
@adamgundry could you please take a look?

optics-core/src/Optics/Label.hs
```diff
@@ -87,9 +106,196 @@
...
+-- Standard Haskell records are a common source of frustration amongst seasoned
+-- Haskell programmers. Their main issues are:
+--
+-- 1) Pollution of global namespace as every field accessor is also a top-level
+--    function.
+--
+-- 2) Inability to define multiple data types sharing field names in the same
+--    module.
+--
+-- 3) Clunky update syntax, especially when nested fields get involved.
```
adamgundry, line 123: Perhaps you could reorder this list so it matches the order in which the points are referenced below?

optics-core/src/Optics/Label.hs
```diff
@@ -87,9 +106,196 @@
...
+-- - <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-DuplicateRecordFields DuplicateRecordFields> - introduced in GHC 8.0.1, addresses (2)
+--
+-- - <https://github.com/ghc-proposals/ghc-proposals/pull/160 NoFieldSelectors> - accepted GHC proposal, addresses (1)
+--
+-- - <https://github.com/ghc-proposals/ghc-proposals/pull/282 RecordDotSyntax> - accepted GHC proposal, addresses (3)
+--
+-- It needs to be noted however that both NoFieldSelectors and RecordDotSyntax
+-- are not yet implemented, with the latter depending on adding setField to
+-- HasField (<https://gitlab.haskell.org/ghc/ghc/issues/16232 ghc/16232>), also not
+-- yet implemented.
```
adamgundry, line 138: Or at least not yet merged :wink: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3257 although that is mostly my fault for not pushing harder on getting feedback and help with the performance issues.

optics-core/src/Optics/Label.hs
```diff
@@ -87,9 +106,196 @@
...
+-- syntax.
+--
+-- == The solution
+--
+-- === Prefixless fields with DuplicateRecordFields
+--
+-- If we look at section showing sample usage, you'll see that field names were
+-- prefixed with full name of a data type, whereas defined labels were
+-- prefixless. This is the simplest approach of using labels as optics that does
+-- not require usage of DuplicateRecordFields, but not the optimal one. We
```
adamgundry, line 153: Should we update the example usage to use `DuplicateRecordFields`, then, if that's the recommended strategy? (And defer the possibility of prefixed field names to some appendix.)

optics-core/src/Optics/Label.hs
```diff
@@ -87,9 +106,196 @@
...
+-- leak into global namespace as accessor functions and can generate a lot of
+-- name clashes. Before NoFieldSelectors is available, this can be alleviated by
+-- splitting modules defining types into two, namely:
 --
--- == Structure of 'LabelOptic' instances
+-- 1) A private one that exports full type definitions, i.e. with their fields
+-- and constructors.
+--
+-- 2) A public one that exports only constructors (or no constructors at all if
+-- the data type in question is opaque).
```
adamgundry, line 215: Hmm, this may be less convenient in a possible future where `setField` and `RecordDotSyntax` are available, because the fields not being in scope will prevent `HasField` constraints being solved (for data abstraction reasons). Though I guess we can cross that bridge if we get to it.

Perhaps we should say something about data abstraction... a consequence of `LabelOptic` (and similar class-based approaches) is that it is not possible to hide the presence of a field for which you have defined a `LabelOptic` instance.

optics-core/src/Optics/Label.hs
arybczak: Makes sense :+1: Done.

optics-core/src/Optics/Label.hs
arybczak: Thanks, I changed it to "not yet merged".

optics-core/src/Optics/Label.hs
arybczak: > Hmm, this may be less convenient in a possible future where setField and RecordDotSyntax are available, because the fields not being in scope will prevent HasField constraints being solved (for data abstraction reasons).

Right, but with RecordDotSyntax (well, NoFieldSelectors really) you won't need this trick, it's just a design pattern (i.e. workaround ;) until these are available.

> Though I guess we can cross that bridge if we get to it.

Agreed.

> Perhaps we should say something about data abstraction... a consequence of LabelOptic (and similar class-based approaches) is that it is not possible to hide the presence of a field for which you have defined a LabelOptic instance.

I added a paragraph that explain the issue below.

optics-core/src/Optics/Label.hs
arybczak: You're right, I found that discrepancy a bit jarring. Updated sample usage section to have prefixless fields as well.

optics-core/src/Optics/Label.hs
```diff
@@ -290,6 +298,8 @@ import Optics.Internal.Optic
 -- prefixed, it works out:
 --
 -- @
+-- {-\# LANGUAGE DuplicateRecordFields #-}
```
arybczak, line 301: This isn't necessary here, right?

arybczak:
> I did an editing pass; I hope you don't mind me being bold and updating this PR. wink

I don't :) Thanks for taking care of these edits.

optics-core/src/Optics/Label.hs
adamgundry: I think you need at least `DisambiguateRecordFields` to allow you to use `name` instead of `User.Internal.name`, don't you? I debated which to add, but went for `DuplicateRecordFields` because it implies `DisambiguateRecordFields` and we're using it elsewhere. We could change this or add a comment with the alternative, though.

arybczak: You don't need it if only one `name` field is in scope. I just checked and you don't even need it if there are multiple `name` fields in scope, which makes sense as once you're inside record syntax for value construction (for update it's a different story I suppose), GHC knows what type you refer to so the choice is unambiguous.

arybczak:
I updated language pragmas.

optics-core/src/Optics/Label.hs
adamgundry: > which makes sense as once you're inside record syntax for value construction (for update it's a different story I suppose), GHC knows what type you refer to so the choice is unambiguous.

Right, this is precisely what `DisambiguateRecordFields` adds over `Haskell2010`: to be able to construct or pattern-match on a record with unqualified fields, if the fields are ambiguous or in scope only qualified. This makes sense because the data constructor tells you which datatype and hence which fields you mean. But it needs (at least) `DisambiguateRecordFields` enabled.

arybczak: Oh, you're right, it doesn't work without any extensions.

I was confused, because I tried it in a project and in worked without `DisambiguateRecordFields` nor `DuplicateRecordFields`. But the project enables `RecordWildCards` globally and it turns out it's enough :thinking:

arybczak:
Looks like all's good now, I'm gonna squash review commits and merge later.

optics-core/src/Optics/Label.hs
adamgundry: Yes, `RecordWildCards` also implies `DisambiguateRecordFields`. Really `DisambiguateRecordFields` should just be part of `GHC2020`... :wink:

Issue #91, opened 2019-01-20
Utilize monoidal structure of (Ix)Traversal and (Ix)Fold


arybczak:
```haskell
foldMempty :: Fold s a
foldMempty = foldVL $ \_ _ -> pure ()

foldMappend
  :: (Is k A_Fold, Is l A_Fold)
  => Optic' k is s a
  -> Optic' l is s a
  -> Fold s a
foldMappend o o' = foldVL $ \f s -> traverseOf_ o f s *> traverseOf_ o' f s
```

```haskell
λ> toListOf (folded % foldMappend _1 (_2 % _Left)) [("how", Left "are"), ("you", Right ())] 
["how","are","you"]
```

It seems like providing instances might be problematic due to types, but utility functions will be nice for things like `has`.

adamgundry:
I suppose we could offer
```hs
instance (k ~ A_Fold) => Monoid (Optic' k is s a)
```
although it has the downside of requiring explicit downcasts when calling `mappend`. I'm ambivalent on the instance, but having named versions of the operations makes sense.

arybczak:
Yeah, explicit downcasts kinda make the instance irrelevant unfortunately. We can surely add regular functions, I'm not sure about good names for them though.

arybczak:
Unfortunately it looks like the full story is a bit more complicated.

```haskell
-- Traversal

oempty :: Monoid t => Traversal s t a b
oempty = traversalVL $ \_ _ -> pure mempty

oplus
  :: (Is k A_Traversal, Is l A_Traversal, Semigroup t)
  => Optic k is s t a b
  -> Optic l js s t a b
  -> Traversal s t a b
oplus o o' = traversalVL $ \f s -> (<>) <$> traverseOf o f s <*> traverseOf o' f s

-- Fold

oempty_ :: Fold s a
oempty_ = foldVL $ \_ _ -> pure ()

oplus_
  :: (Is k A_Fold, Is l A_Fold)
  => Optic' k is s a
  -> Optic' l js s a
  -> Fold s a
oplus_ o o' = foldVL $ \f s -> traverseOf_ o f s *> traverseOf_ o' f s

-- Indexed traversal

ioempty :: Monoid t => IxTraversal i s t a b
ioempty = ixTraversalVL $ \_ _ -> pure mempty

ioplus
  :: (Is k A_Traversal, Is l A_Traversal, Semigroup t)
  => Optic k (WithIx i) s t a b
  -> Optic l (WithIx i) s t a b
[...]

PR #327, opened 2020-07-17
Monoidal concatenation

serras:
The idea is to generalize the `summing` operation in `Data.Fold`, so that it give stronger guarantees. For example, if you sum two traversals, you can get a traversal out of it.

I have also included an example of usage to return elements from a `Tree` on increasing level.

### DIFF
--- a/optics-core/src/Optics/Lens.hs
+++ b/optics-core/src/Optics/Lens.hs
@@ -232,7 +232,7 @@ split
 split l r = withLens l $ \getl setl ->
                 withLens r $ \getr setr ->
   lens (\s -> (getl s, getr s))
-       (\s (l, r) -> setr (setl s l) r)
+       (\s (a, b) -> setr (setl s a) b)
 {-# INLINE split #-}
 
 -- | We can always retrieve a @()@ from any type.
--- a/optics-core/src/Optics/Lens.hs
+++ b/optics-core/src/Optics/Lens.hs
@@ -93,6 +93,7 @@ module Optics.Lens
   , equality'
   , chosen
   , alongside
+  , split
   , united
 
   -- * Additional elimination forms
@@ -216,6 +217,23 @@ alongside l r = withLens l $ \getl setl ->
   lens (\(s, s')         -> (getl s,   getr s'   ))
        (\(s, s') (b, b') -> (setl s b, setr s' b'))
 {-# INLINE alongside #-}
+
+-- | Make a 'Lens' from two other lenses by executing them on
+-- the same value. The lenses should focus on different parts
+-- of the value to be well-behaved.
+--
+-- >>> ('a', 'b') ^. split _1 _2
+-- ('a','b')
+split
+  :: (Is k A_Lens, Is l A_Lens)
+  => Optic k is s s a  b
+  -> Optic l js s s a' b'
+  -> Lens s s (a, a') (b, b')
+split l r = withLens l $ \getl setl ->
+                withLens r $ \getr setr ->
+  lens (\s -> (getl s, getr s))
+       (\s (l, r) -> setr (setl s l) r)
+{-# INLINE split #-}
 
 -- | We can always retrieve a @()@ from any type.
 --
--- a/optics-core/src/Optics/Monoidal.hs
+++ b/optics-core/src/Optics/Monoidal.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TypeInType #-}
 {-# LANGUAGE UndecidableInstances #-}
 module Optics.Monoidal (
   After(..)
--- a/optics-core/src/Optics/Monoidal.hs
+++ b/optics-core/src/Optics/Monoidal.hs
@@ -1,0 +1,193 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE UndecidableInstances #-}
+module Optics.Monoidal where
+
+import GHC.TypeLits
+import Optics.Fold
+import Optics.IxFold
+import Optics.IxSetter
+import Optics.IxTraversal
+import Optics.Internal.Optic.Types
+import Optics.Optic
+import Optics.Setter
+import Optics.Traversal
+
+-- Essentially this collapses the hierarchy into
+-- Traversal, Fold, and Setter,
+-- and then takes the usual Join
+type family AfterJoin (k :: OpticKind) (l :: OpticKind) where
+  -- BEGIN GENERATED CONTENT
+  -- An_Iso-----
+  -- no AfterJoin with         A_ReversedLens
+  AfterJoin An_Iso             A_ReversedPrism    = A_Fold
+  AfterJoin An_Iso             A_Prism            = A_Traversal
+  -- no AfterJoin with         A_Review
+  AfterJoin An_Iso             A_Lens             = A_Traversal
+  AfterJoin An_Iso             A_Getter           = A_Fold
+  AfterJoin An_Iso             An_AffineTraversal = A_Traversal
+  AfterJoin An_Iso             An_AffineFold      = A_Fold
+  AfterJoin An_Iso             A_Traversal        = A_Traversal
+  AfterJoin An_Iso             A_Fold             = A_Fold
+  AfterJoin An_Iso             A_Setter           = A_Setter
+
+  -- no AfterJoin with A_ReversedLens-----
+
+  -- A_ReversedPrism-----
+  AfterJoin A_ReversedPrism    An_Iso             = A_Fold
+  -- no AfterJoin with         A_ReversedLens
+  AfterJoin A_ReversedPrism    A_Prism            = A_Fold
+  -- no AfterJoin with         A_Review
+  AfterJoin A_ReversedPrism    A_Lens             = A_Fold
+  AfterJoin A_ReversedPrism    A_Getter           = A_Fold
+  AfterJoin A_ReversedPrism    An_AffineTraversal = A_Fold
+  AfterJoin A_ReversedPrism    An_AffineFold      = A_Fold
+  AfterJoin A_ReversedPrism    A_Traversal        = A_Fold
+  AfterJoin A_ReversedPrism    A_Fold             = A_Fold
+  -- no AfterJoin with         A_Setter
+
+  -- A_Prism-----
+  AfterJoin A_Prism            An_Iso             = A_Traversal
+  -- no AfterJoin with         A_ReversedLens
+  AfterJoin A_Prism            A_ReversedPrism    = A_Fold
+  -- no AfterJoin with         A_Review
+  AfterJoin A_Prism            A_Lens             = A_Traversal
+  AfterJoin A_Prism            A_Getter           = A_Fold
+  AfterJoin A_Prism            An_AffineTraversal = A_Traversal
+  AfterJoin A_Prism            An_AffineFold      = A_Fold
+  AfterJoin A_Prism            A_Traversal        = A_Traversal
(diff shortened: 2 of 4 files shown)


optics-core/src/Optics/Monoidal.hs
```diff
@@ -0,0 +1,193 @@
...
+import Optics.IxTraversal
+import Optics.Internal.Optic.Types
+import Optics.Optic
+import Optics.Setter
+import Optics.Traversal
+
+-- Essentially this collapses the hierarchy into
+-- Traversal, Fold, and Setter,
+-- and then takes the usual Join
+type family AfterJoin (k :: OpticKind) (l :: OpticKind) where
```
adamgundry, line 20: Perhaps I'm missing something, but is it the case that `AfterJoin k l ~ Join A_Traversal (Join k l)`? Couldn't we use that definition instead?

optics-core/src/Optics/Monoidal.hs
```diff
@@ -0,0 +1,193 @@
...
+
+  -- Everything else is a type error.
+  AfterJoin k l = TypeError ('ShowType k
+                       ':<>: 'Text " and "
+                       ':<>: 'ShowType l
+                       ':<>: 'Text " cannot be composed sequentially")
+
+infixr 6 <++>
+class After is m where
+  nothing :: Optic' m is s a
```
adamgundry, line 169: We already have `ignored :: IxAffineTraversal i s s a b` which is strictly more general than `nothing`, isn't it?

optics-core/src/Optics/Monoidal.hs
```diff
@@ -0,0 +1,193 @@
...
+  (<++>)  :: (Is k m, Is l m, m ~ AfterJoin k l)
+          => Optic' k is s a -> Optic' l is s a
+          -> Optic' m is s a
```
adamgundry, line 172: In general, we tend to avoid introducing new operators in `optics`, or at least provide a non-operator equivalent. Is there precedent for this operator elsewhere? I suppose `summing` would be the obvious name choice if we don't mind a slightly breaking change.

optics-core/src/Optics/Monoidal.hs
```diff
@@ -0,0 +1,193 @@
...
+
+instance After NoIx A_Traversal where
+  nothing = traversalVL $ \_ s -> pure s
+  t1 <++> t2 = traversalVL $ \f s -> traverseOf t1 f s *> traverseOf t2 f s
```
adamgundry, line 176: I wonder if there is any benefit to implementing this directly on the profunctor representation rather than going via the VL representation...

phadej, line 176: The `*>` is suspicious. What `Setter` made from this traversal does. (I suspect nothing to the first half).

Yet, the `After NoIx A_Setter` instance does something else.

This feels like a giving shotgun to a kid.

optics-core/src/Optics/Monoidal.hs
```diff
@@ -0,0 +1,193 @@
...
+instance After NoIx A_Fold where
+  nothing = foldring (\_ r _ -> r)
+  (<++>) = summing
+instance After (WithIx ix) A_Fold where
+  nothing = ifoldring (\_ r _ -> r)
+  (<++>) = isumming
+
+instance After NoIx A_Setter where
+  nothing = sets (\_ s -> s)
+  s1 <++> s2 = sets (\f s -> over s2 f (over s1 f s))
```
phadej, line 190: There should be a big warning. The resulting `Setter` is lawful only(?) if halves of it don't modify the same parts of underlying structure.

I think that in this case, `<++>` have to be commutative operator. Yet for folds that doesn't make sense.

This structure feels very ad-hoc to me. I'm not sure this is good idea at all to overload this (or `summing` or ...) name.

optics-core/src/Optics/Monoidal.hs
serras: I haven't thought deeply about it, but I guess that would make the trick, since `A_Traversal` is the "bottom" of that restricted hierarchy.

optics-core/src/Optics/Monoidal.hs
serras: Not really, if only because of the types. `nothing` here can be a `Setter`, `Fold`, or `Traversal`, everything more general than `ignored`.

optics-core/src/Optics/Monoidal.hs
serras: I didn't know about that rule. I haven't seen this operator used anywhere else, but I think that `summing` would work great here (and would provide a good backwards-compatibility story for those already using `summing` in `Fold`).

optics-core/src/Optics/Monoidal.hs
serras: Right, I see that there is a discrepancy between what the instances for `Setter` and `Traversal` do.

optics-core/src/Optics/Monoidal.hs
serras: That is exactly the use case I had in mind. Imagine that I want to define several `Traversal`s which works over binary trees. With this API I could define it as:

```haskell
data BinaryTree a = Leaf { _leaf :: a } | Node { _left, _right :: BinaryTree a }

preOrder = leaf <++> left % preOrder <++> right % preOrder
inOrder = left % preOrder <++> leaf <++> right % preOrder
postOrder = left % preOrder <++> right % preOrder <++> leaf
```

optics-core/src/Optics/Monoidal.hs
phadej: AffineTraversals are (i.e. can be cast to) Setters, Folds and Traversals.

adamgundry: Right, I mean `ignored` is more general in terms of the `Is` relation. It doesn't strictly have a more general type, but `optics`-style polymorphism will allow it to be used with eliminators for setters, folds and traverals.

adamgundry:
@phadej thanks for catching the Traversal/Setter difference and the unlawfulness. I buy that `adjoin` is potentially unlawful for non-disjoint Traversals (and I agree with your comments wrt "unsound" being too strong). But is there an example of breaking the laws for Setter alone? I can't immediately see one, because Setter doesn't give you a way to observe how many times the update function is applied.

optics-core/src/Optics/Monoidal.hs
adamgundry: This is a nice example use case, and could do with appearing in the documentation, whatever API variant we end up with. :smile:

phadej:
@adamgundry you can break `over setter f . over setter g = over setter (f . g)` law quite easily.

```haskell
Prelude Optics Optics.Monoidal Data.Char> let setter = castOptic @A_Setter simple <++> castOptic @A_Setter simple
Prelude Optics Optics.Monoidal Data.Char> over setter (+1) . over setter (*2) $ (3 :: Int)
14
Prelude Optics Optics.Monoidal Data.Char> over setter ((+1) . (*2)) $ (3 :: Int)
15
```

arybczak:
This is similar to what I tried to do in https://github.com/well-typed/optics/issues/91#issuecomment-460946013

Combination of results with read-only optics works out because you can discard the structure you traverse, so `*>` does the job fine, but the setter part needs to preserve the structure of traversed data and I suspect it cannot be done in general, so it simply won't work for things that are a Setter.

What `incrLevels` does when used as a traversal? I suspect it doesn't work as advertised.

optics-core/src/Optics/Monoidal.hs
arybczak: `castOptic ignored` (and `noIx (castOptic ignored)` when explicit unindexed variant is needed) do the job.

serras:
I have just pushed a new attempt, which satisfies the traversal laws if the traversals focus on different parts of the value.

```haskell
*Optics Optics.Monoidal Data.Char> data BinaryTree a = Leaf { _leaf :: a } | Node { _left, _right :: BinaryTree a } deriving Show
*Optics Optics.Monoidal Data.Char> makeLenses ''BinaryTree;
*Optics Optics.Monoidal Data.Char> let preOrder :: Traversal' (BinaryTree a) a; preOrder = leaf <++> (left % preOrder) <++> (right % preOrder)
*Optics Optics.Monoidal Data.Char> toListOf preOrder $ Node (Leaf 'x') (Node (Leaf 'y') (Leaf 'z'))
"xyz"
*Optics Optics.Monoidal Data.Char> over preOrder toUpper $ Node (Leaf 'x') (Node (Leaf 'y') (Leaf 'z'))
Node {_left = Leaf {_leaf = 'X'}, _right = Node {_left = Leaf {_leaf = 'Y'}, _right = Leaf {_leaf = 'Z'}}}
```

serras:
Note that my end goal is also to provide a generalized interface to `failing`, which does not only work on `Fold`. So for example you could compose two affine traversals and that would focus on the first one that matches.

optics-core/src/Optics/Lens.hs
```diff
@@ -217,6 +218,23 @@ alongside l r = withLens l $ \getl setl ->
        (\(s, s') (b, b') -> (setl s b, setr s' b'))
 {-# INLINE alongside #-}
 
+-- | Make a 'Lens' from two other lenses by executing them on
+-- the same value. The lenses should focus on different parts
+-- of the value to be well-behaved.
+--
+-- >>> ('a', 'b') ^. split _1 _2
+-- ('a','b')
+split
```
arybczak, line 227: Why not make it an instance of `After`?

arybczak:
I'm not convinced that introducing a type class for this is a good idea. The only non-trivial instances currently are for `Traversal` and `Setter`, both of which come with a set of caveats.

Since `split` and `After` instance for `Traversal` are already in `lens` under `lensProduct` and `adjoin`, I'd be inclined to include them under these names in `optics` to avoid introduction of unnecessary api differences.

arybczak:
After some experimentation I got this:

```haskell

adjoin
  :: (Is k A_Traversal, Is l A_Traversal)
  => Optic' k is s a
  -> Optic' l js s a
  -> Traversal' s a
adjoin o1 o2 = combined % traversed
  where
    combined = lensVL $ \f s -> evalState (    traverseOf o1 update s
                                           >>= traverseOf o2 update)
      <$> f (   toListOf (castOptic @A_Traversal o1) s
             ++ toListOf (castOptic @A_Traversal o2) s)

    update a = get >>= \case
      a' : as' -> put as' >> pure a'
      []       ->            pure a
{-# INLINE adjoin #-}

-- indexed variant
iadjoin
  :: (Is k A_Traversal, Is l A_Traversal, is `HasSingleIndex` i)
  => Optic' k is s a
  -> Optic' l is s a
  -> IxTraversal' i s a
iadjoin o1 o2 = conjoined (adjoin o1 o2) (combined % traversed % itraversed)
  where
    combined = lensVL $ \f s -> evalState (    traverseOf o1 update s
                                           >>= traverseOf o2 update)
      <$> f (   itoListOf (castOptic @A_Traversal o1) s
             ++ itoListOf (castOptic @A_Traversal o2) s)

    update a = get >>= \case
      (_, a') : as' -> put as' >> pure a'
      []            ->            pure a
{-# INLINE iadjoin #-}
```

arybczak:
One more argument against the type class: things like `_1 <++> _2` don't type check, whereas `adjoin _1 _2` does, because `adjoin` accepts anything that's a traversal, whereas `<++>` accepts a pair of traversals or a pair of folds etc.

That's the reason we have `summing` and `asumming` instead of having `summing` as a class method.

adamgundry:
@arybczak thanks for looking at this. I'd be happy with `lensProduct` and `adjoin` for lens compatibility. I don't have a strong sense of whether it is worth offering a version for `Setter`.

We should have "Monoid structure" sections of the `Optics.[Ix](Setter|Traversal)` docs, and perhaps an `Optics.Monoidal` module that documents all these monoids, whether or not it actually exposes them as a type class.

serras:
Given this discussion, I am going to close the PR, since clearly is should not be merged as is. Let me know if I could help bringing the other operators to `optics`, as I find them quite useful, even though they might not be "safe" unless you point to different parts of a value.

PR #330, opened 2020-07-30
Leverage HasField to obtain field getters

arybczak:
I wonder if it makes sense to do that instead of reporting an error about lack of instance.

I'm tentatively positive about this though:

```haskell
λ> :set -XOverloadedLabels 
λ> data User = User { name :: String, age :: Int }
λ> let user = User "Tom" 23
λ> user ^. #name
"Tom"
λ> user ^. #namey

<interactive>:7:1: error:
    • No instance for (GHC.Records.HasField "namey" User ())
        arising from a use of ‘it’
    • In the first argument of ‘print’, namely ‘it’
      In a stmt of an interactive GHCi command: print it
λ> user & #name .~ "Bob"

<interactive>:6:8: error:
    • A_Getter cannot be used as A_Setter
      Perhaps you meant one of these:
        ‘view’ (from Optics.Getter)
        ‘(^.)’ (from Optics.Operators)
    • In the second argument of ‘(&)’, namely ‘#name .~ "Bob"’
      In the expression: user & #name .~ "Bob"
      In an equation for ‘it’: it = user & #name .~ "Bob"
```

### DIFF
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -43,10 +43,12 @@ import Data.Proxy (Proxy (..))
 import Data.Type.Equality
 import GHC.Generics (Rep)
 import GHC.OverloadedLabels
+import GHC.Records
 import GHC.TypeLits
 
 import Data.Profunctor.Indexed
 
+import Optics.Internal.Bi
 import Optics.Internal.Optic.Subtyping
 import Optics.Internal.Optic.TypeLevel
 import Optics.Internal.Optic.Types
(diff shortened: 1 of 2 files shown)


arybczak:
Looks like 8.2 is needed for this.

arybczak:
@adamgundry what do you think?

adamgundry:
I forgot to mention: I'm okay with dropping 8.0 support, assuming we decide to go for this.

adamgundry:
That's understandable, but disappointing. I guess it would be less of a problem if we could make a lens rather than just a getter, so I should keep trying to land `setField`...

I can't currently see a way to wait until we know `HasField` before committing to `k ~ A_Getter`, except using generics.

arybczak:
Closing for now as the type error in case of misspelled setter isn't helpful. Let's wait for `setField` :)

PR #332, opened 2020-07-31
Add (i)adjoin

### LINKED ISSUE #331: Add lensProduct and adjoin
See #327, but opening an issue so we don't forget about this. I've wanted `adjoin` in the past for `lens` and not been able to find it, so I think we should have these and document them properly.

arybczak:
Fixes #331.

It's reasonably fast when compared to a hand-written version (and obliterates `lens`):

```
benchmarking misc/adjoin/hand-written
time                 30.16 ns   (29.32 ns .. 31.73 ns)
                     0.991 R²   (0.978 R² .. 1.000 R²)
mean                 30.34 ns   (29.49 ns .. 32.66 ns)
std dev              2.938 ns   (876.6 ps .. 5.151 ns)
variance introduced by outliers: 90% (severely inflated)

benchmarking misc/adjoin/adjoin
time                 65.72 ns   (65.14 ns .. 66.25 ns)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 65.72 ns   (65.10 ns .. 66.22 ns)
std dev              1.131 ns   (853.5 ps .. 1.562 ns)
variance introduced by outliers: 21% (moderately inflated)

benchmarking misc/adjoin/adjoin/lens
time                 3.952 μs   (3.911 μs .. 4.019 μs)
                     0.997 R²   (0.991 R² .. 1.000 R²)
mean                 4.016 μs   (3.930 μs .. 4.307 μs)
std dev              309.4 ns   (65.82 ns .. 594.4 ns)
variance introduced by outliers: 78% (severely inflated)
```

When adjoining traversals that overlap I'm not sure which laws might be violated though. I suspect traversal composition law can be violated when the traversing function is stateful. I don't think setter laws can be violated.

Because of this I'm for exporting it normally.

### DIFF
--- a/optics-core/src/Optics/IxTraversal.hs
+++ b/optics-core/src/Optics/IxTraversal.hs
@@ -59,6 +59,7 @@ module Optics.IxTraversal
   , ibackwards
   , ipartsOf
   , isingular
+  , iadjoin
 
   -- * Subtyping
   , A_Traversal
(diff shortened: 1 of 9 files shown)


phadej:
The violated "law" is "5.5  No duplication of elements" in http://www.cs.ox.ac.uk/jeremy.gibbons/publications/iterator.pdf

arybczak:
@phadej I see. I don't get why is that a problem though.

```haskell
λ> let once = simple
λ> let twice = simple `adjoin` simple
λ> over once (+1) 7
8
λ> over twice (+1) 7
8
```

Even though illegal traverses the value twice, it changes it only once, because the second time it's still given the original value, not the one that was just modified (of course if you use `traverseOf` instead of `over` with a stateful traversal you can most likely distinguish between them, but it's not obvious).

adamgundry:
This looks great to me. I agree that exporting it normally is fine; it's no worse than say `partsOf`. I'd (still) suggest expanding on the docs slightly (https://github.com/well-typed/optics/pull/327#issuecomment-665095162):
> We should have "Monoid structure" sections of the `Optics.[Ix](Setter|Traversal)` docs, and perhaps an `Optics.Monoidal` module that documents all these monoids

although maybe a section in `Optic` would be preferable to a new module. I can draft it if you like.

arybczak:
> I can draft it if you like.

Yes please :pray:

optics-core/src/Optics/IxTraversal.hs
```diff
@@ -340,14 +356,24 @@ isingular o = conjoined (singular o) $ iatraversalVL $ \point f s ->
...
--- For the fold version see 'Optics.IxFold.isumming'.
+-- /Note:/ if the argument traversals are not disjoint, the result will not
+-- respect the 'IxTraversal' laws, because it will visit the same element multiple
+-- times.  See section 7 of
+-- <https://www.cs.ox.ac.uk/jeremy.gibbons/publications/uitbaf.pdf Understanding Idiomatic Traversals Backwards and Forwards>
+-- by Bird et al. for why this is illegal.
+--
+-- >>> iview (ipartsOf (each `iadjoin` each)) ("x","y")
+-- ([0,1,0,1],["x","y","x","y"])
+-- >>> iset (ipartsOf (each `iadjoin` each)) (const ["a","b","c","d"]) ("x","y")
```
arybczak, line 372: Nice example :+1:

PR #335, opened 2020-08-06
Make chosen an indexed lens to see which value was traversed

arybczak:


arybczak:
I added a reference to `chosen` in `Additional introduction forms` of `Optics.Lens`.

PR #340, opened 2020-08-18
Test with GHC 8.8.4 and 8.10.2

arybczak:


adamgundry:
@arybczak Thanks for taking care of this!

Issue #338, opened 2020-08-17
optics-th's test suite failed to build with GHC 8.10.2


felixonmars:
```
Building test suite 'optics-th-tests' for optics-th-0.3.0.1..
[1 of 3] Compiling Optics.TH.Tests.DuplicateRecordFields ( tests/Optics/TH/Tests/DuplicateRecordFields.hs, dist/build/optics-th-tests/optics-th-tests-tmp/Optics/TH/Tests/DuplicateRecordFields.dyn_o )

tests/Optics/TH/Tests/DuplicateRecordFields.hs:37:34: error:
    • Multiple declarations of ‘foo’
      Declared at: tests/Optics/TH/Tests/DuplicateRecordFields.hs:20:3
                   tests/Optics/TH/Tests/DuplicateRecordFields.hs:37:34
    • In the Template Haskell quotation
        [d| data family F x
            data instance F Int = MkF {foo :: Int} |]
   |
37 |      data instance F Int = MkF { foo :: Int }|])
   |                                  ^^^
```

arybczak:
Indeed, I noticed that yesterday. Looks like GHC bug, it works with 8.10.1.

alexeyzab:
I was about to mention it too, I made it so that we expect the compilation failure on [stackage](https://github.com/commercialhaskell/stackage/issues/5606) for now.

adamgundry:
Thanks for the report! Ironically I think the culprit is actually a fix to a GHC bug, namely https://gitlab.haskell.org/ghc/ghc/-/issues/17965. The current version of the test shouldn't compile after all, because it uses `foo` as both a field label and a top-level identifier, and that isn't supposed to be allowed. Trivial fix incoming.

arybczak:
Thanks @adamgundry for taking care of it. I think the fix warrants 0.3.0.2 release of optics-th.

PR #341, opened 2020-08-18
Bump version of optics-th to 0.3.0.2

arybczak:
Related to #338.

I wonder how we should tag releases. `v0.3.0.2` would be somewhat ambiguous here.


arybczak:
Setting `buildable: False` would work as it prevents cabal from building the bit, but looks like Hackage disallows it in metadata revisions (I just tried to add it and it wouldn't let me).

arybczak:
We can deprecate 0.3.0.1 though.

PR #349, opened 2020-08-30
Fix cabal-version for template-haskell-optics

adamgundry:
Currently `cabal check` objects due to the presence of `extra-doc-files`.


arybczak:
Ah I see, I added `extra-doc-files` after the release (once I noticed that `changelog` wasn't included), that's why Hackage didn't complain. Thanks for spotting that :+1:

phadej:
@arybczak never push to `master` directly...

PR #352, opened 2020-09-06
Print missing language extensions on TH generation of labels

### LINKED ISSUE #342: Better docs/errors for missing extensions with TH declaration splices
When using TH declaration splices such as `makeFieldLabels`, various language extensions must be enabled, otherwise the spliced code will not type-check (see https://github.com/well-typed/optics/pull/189#issuecomment-500731253).

We should include the list of required extensions in the documentation, but better yet, the TH code itself should check they are enabled and report a nice error message if not (e.g. `persistent-template` does this: https://github.com/yesodweb/persistent/blob/10f689371345ec20edf0d7e8d776f7f4bc4a187b/persistent-template/Database/Persist/TH.hs#L1895).

arybczak:
Fixes #342.

```haskell
unknown@electronics optics $ cabal repl optics-th
Build profile: -w ghc-8.10.2 -O1
In order, the following will be built (use -v for more details):
 - optics-th-0.3.0.2 (lib) (file src/Optics/TH/Internal/Product.hs changed)
Preprocessing library for optics-th-0.3.0.2..
GHCi, version 8.10.2: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/unknown/.ghci
[1 of 5] Compiling Language.Haskell.TH.Optics.Internal ( src/Language/Haskell/TH/Optics/Internal.hs, interpreted )
[2 of 5] Compiling Optics.TH.Internal.Utils ( src/Optics/TH/Internal/Utils.hs, interpreted )
[3 of 5] Compiling Optics.TH.Internal.Sum ( src/Optics/TH/Internal/Sum.hs, interpreted )
[4 of 5] Compiling Optics.TH.Internal.Product ( src/Optics/TH/Internal/Product.hs, interpreted )
[5 of 5] Compiling Optics.TH        ( src/Optics/TH.hs, interpreted )
Ok, five modules loaded.
λ> :set -XTemplateHaskell
λ> data X = X { xX :: Int };
λ> ; makeFieldLabels ''X;

<interactive>:3:3: error:
    Generating LabelOptic instances requires the following language extensions:

- DataKinds
- FlexibleInstances
- MultiParamTypeClasses
- TypeFamilies
- UndecidableInstances

Please enable the extensions by copy/pasting these lines into the top of your file:

{-# LANGUAGE DataKinds #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE TypeFamilies #-}

To enable them in the GHCi session, use the following command:

:set -XDataKinds -XFlexibleInstances -XMultiParamTypeClasses -XTypeFamilies -XUndecidableInstances
```

### DIFF
--- a/optics-th/src/Optics/TH/Internal/Utils.hs
+++ b/optics-th/src/Optics/TH/Internal/Utils.hs
@@ -105,8 +105,8 @@ requireExtensions what extLists = do
       , intercalate "\n" (map (("- " ++) . show) extensions)
       , "\n\nPlease enable the extensions by copy/pasting these lines into the top of your file:\n\n"
       , intercalate "\n" (map extensionToPragma extensions)
-      , "\n\nTo enable them in the GHCi session, use the following command:\n\n"
-      , ":set " ++ unwords (map (("-X" ++) . show) extensions)
+      , "\n\nTo enable them in a GHCi session, use the following command:\n\n"
+      , ":seti " ++ unwords (map (("-X" ++) . show) extensions)
       ]
   where
     extensionToPragma ext = "{-# LANGUAGE " ++ show ext ++ " #-}"
--- a/optics-th/src/Optics/TH/Internal/Utils.hs
+++ b/optics-th/src/Optics/TH/Internal/Utils.hs
@@ -1,6 +1,8 @@
 module Optics.TH.Internal.Utils where
 
+import Control.Monad
 import Data.Maybe
+import Data.List
 import Language.Haskell.TH
 import qualified Data.Map as M
 import qualified Data.Set as S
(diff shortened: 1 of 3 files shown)


optics-th/src/Optics/TH/Internal/Utils.hs
```diff
@@ -76,6 +78,48 @@ quantifyType' exclude vars cx t = ForallT vs cx t
...
+requireExtensions what extLists = do
+  -- Taken from the persistent library
+  required <- filterM (fmap (not . or) . traverse isExtEnabled) extLists
+  case mapMaybe listToMaybe required of
+    [] -> pure ()
+    [extension] -> fail $ mconcat
+      [ "Generating " ++ what ++ " requires the "
+      , show extension
+      , " language extension. Please enable it by copy/pasting this line to the top of your file:\n\n"
+      , extensionToPragma extension
```
adamgundry, line 101: I like your addition below about the GHCi session. Might it be worth including for the single-extension case as well?

optics-th/src/Optics/TH/Internal/Utils.hs
```diff
@@ -76,6 +78,48 @@ quantifyType' exclude vars cx t = ForallT vs cx t
...
+      ]
+    extensions -> fail $ mconcat
+      [ "Generating " ++ what ++ " requires the following language extensions:\n\n"
+      , intercalate "\n" (map (("- " ++) . show) extensions)
+      , "\n\nPlease enable the extensions by copy/pasting these lines into the top of your file:\n\n"
+      , intercalate "\n" (map extensionToPragma extensions)
+      , "\n\nTo enable them in the GHCi session, use the following command:\n\n"
+      , ":set " ++ unwords (map (("-X" ++) . show) extensions)
```
adamgundry, line 109:
```suggestion
      , "\n\nTo enable them in a GHCi session, use the following command:\n\n"
      , ":seti " ++ unwords (map (("-X" ++) . show) extensions)
```

adamgundry (approved):
@arybczak Lovely! Thanks for taking care of this. A couple of minor suggestions in addition to the review comments:
 * `makeFields` could have a `requireExtensions` call.
 * I'm slightly tempted to include the list of required extensions explicitly in the haddocks.

arybczak:
> makeFields could have a requireExtensions call

I wanted to do that, the problem is that its need for `FunctionalDependencies` depends on whether the `HasX` class will be generated.

arybczak:
> I'm slightly tempted to include the list of required extensions explicitly in the haddocks.

There are 7 variations for functions that generate labels (`makeFieldLabels`, `makeFieldLabelsFor`, `makeFieldLabelsWith`, `declareFieldLabels`, `declareFieldLabelsFor`, `declareFieldLabelsWith`, `makePrismLabels`), so I'd skip it.

phadej:
`FunctionalDependencies` is "just" a better `MultiParamTypeClasses`. Yes, HLint (and maybe some other tools) will say that you don't need `FunDep` is not needed, as `MPTC` is sufficient, but they are wrong. IMO better just to `FunDep`, after all `LabelOptic` class uses fundeps, so it's not like you ask for something extra.

arybczak:
Fair enough, I added check for `FunctionalDependencies`.

arybczak:
Though `makeFields` also needs `FlexibleInstances` for higher rank fields :roll_eyes: Sigh.

optics-th/src/Optics/TH/Internal/Utils.hs
arybczak: Yeah, I forgot about that, done.

arybczak:
I made it require both. HLint will not complain about redundant pragmas because it can't inspect TH.

phadej:
>  HLint will not complain about redundant pragmas because it can't inspect TH.

It may for users who copy-paste the `LANGUAGE` pragma block. (But it's ok to me).

adamgundry:
I'm okay with this. A slight over-approximation to the required extensions seems relatively unproblematic, although it might conceivably result in users needing to enable additional extensions on a working build. And fair enough regarding docs.

PR #359, opened 2020-10-04
Leverage HasField to obtain field lenses in GHC >= 9.1

arybczak:
For when https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3257 is merged ;)

### DIFF
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -41,6 +41,9 @@ import Data.Function ((&))
 import Data.Type.Equality
 import GHC.Generics (Rep)
 import GHC.OverloadedLabels
+#if __GLASGOW_HASKELL__ >= 901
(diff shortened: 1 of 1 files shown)


adamgundry:
You beat me to it. :smile: 

I had wondered if it is worth keeping `GeneralLabelOptic`, given this change. I suppose that (for the foreseeable future) `HasField` doesn't let us generate optic kinds other than lenses, whereas generics do, so we might as well keep the option to override it.

When we actually do this, we'll need to clearly document how it works, especially because it is a GHC-version-dependent API. And the haddocks may be built with a different GHC version.

arybczak:
> I had wondered if it is worth keeping `GeneralLabelOptic`, given this change.

Yeah, I'm on the fence with that. The incoherence can be somewhat confusing.

arybczak:
Closing for now as the plan is to split the `HasField` into two classes.

PR #366, opened 2020-10-19
RFC: Make one class instead of Field1..

phadej:
I only changed `Field1`, but others are easy to do too..

### DIFF
--- a/optics-core/src/Data/Tuple/Optics.hs
+++ b/optics-core/src/Data/Tuple/Optics.hs
@@ -24,11 +24,15 @@
 -- >>> set _2 'x' (MkT 1 False)
 -- MkT 1 'x'
 --
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE UndecidableInstances   #-}
 module Data.Tuple.Optics
   (
+    PositionOptic (..)
   -- * Tuples
-    Field1(..)
+  , Field1, _1
   , Field2(..)
   , Field3(..)
   , Field4(..)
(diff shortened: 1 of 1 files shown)


optics-core/src/Data/Tuple/Optics.hs
```diff
@@ -46,78 +50,91 @@ import Data.Functor.Product
...
+-- ((),"world")
+--
+-- This can also be used on larger tuples as well:
+--
+-- >>> (1,2,3,4,5) & _1 %~ (+41)
+-- (42,2,3,4,5)
+_1 :: Field1 s t a b => Lens s t a b
+_1 = positionOptic @0
+
+instance PositionOptic 0 An_Iso (Identity a) (Identity b) a b where
```
phadej, line 95: Note: `_1` for `Identity` is `An_Iso`.

In theory one can have `_1` and `_2` to be`Prisms` for `Either a b`... whether that is a good idea, I'm not sure :)

optics-core/src/Data/Tuple/Optics.hs
```diff
@@ -46,78 +50,91 @@ import Data.Functor.Product
...
+-- >>> traverseOf _1 putStrLn ("hello","world")
+-- hello
+-- ((),"world")
+--
+-- This can also be used on larger tuples as well:
+--
+-- >>> (1,2,3,4,5) & _1 %~ (+41)
+-- (42,2,3,4,5)
+_1 :: Field1 s t a b => Lens s t a b
+_1 = positionOptic @0
```
phadej, line 93: there should be `{-# INLINE _1 -}`, I guess.

phadej:
ping

optics-core/src/Data/Tuple/Optics.hs
adamgundry: I would be inclined to keep `_1` as always a `Lens` in the interests of simplicity. It's one less parameter to infer (perhaps wrongly), and I think code using it as a prism would be just confusing! I suppose `PositionOptic` could be polymorphic, then one could reuse it for `Prism`s with a different function (e.g. `con1 :: PositionOptic 0 A_Prism s t a b -> Prism s t a b`).

optics-core/src/Data/Tuple/Optics.hs
```diff
@@ -46,78 +50,91 @@ import Data.Functor.Product
...
-  -- This can also be used on larger tuples as well:
-  --
-  -- >>> (1,2,3,4,5) & _1 %~ (+41)
-  -- (42,2,3,4,5)
-  _1 :: Lens s t a b
-  default _1 :: (Generic s, Generic t, GIxed N0 (Rep s) (Rep t) a b)
-             => Lens s t a b
-  _1 = ix proxyN0
-  {-# INLINE[1] _1 #-}
+type Field1 = PositionOptic 0 A_Lens
```
adamgundry, line 74: I wonder about the impact on type error messages. Presumably this type synonym will tend to get expanded, and the off-by-one is mildly confusing.

adamgundry (commented):
Sorry for the delay in getting to this.

I'm a little unsure of the motivation here. Having one class rather than N is certainly tidier, but do you anticipate being able to define any interesting `PositionOptic` instances that are polymorphic in the `pos` parameter? (For example, I guess we could make `_n` work like `ix n` if we really wanted to, but I somehow doubt we do...) Are there other good reasons than reducing boilerplate?

My concern is that this will make inferred types / errors slightly more confusing, for comparatively little benefit. But I may well be missing things, and I'm not strongly opposed.

arybczak:
I have similar thoughts as @adamgundry - I'm not sure what the motivation is here.

Btw, this PR has some overlap with #361 and there positions (with `gposition`) start with `1`.

Issue #322, opened 2020-06-04
Tests for type errors


adamgundry:
It occurred to me that the test suite could do with a more thorough set of tests for the type errors that are generated by invalid code, to prevent error message regressions. We could either do this with doctests, or use `-fdefer-type-errors` and catch and inspect type errors at runtime.

phadej:
It's tricky.

```
*Optics.Core> :set -fdefer-type-errors 
*Optics.Core> let x = 'x' + True

<interactive>:2:15: warning: [-Wdeferred-type-errors]
    • Couldn't match expected type ‘Char’ with actual type ‘Bool’
    • In the second argument of ‘(+)’, namely ‘True’
      In the expression: 'x' + True
      In an equation for ‘x’: x = 'x' + True
*Optics.Core> x
*** Exception: <interactive>:2:15: error:
    • Couldn't match expected type ‘Char’ with actual type ‘Bool’
    • In the second argument of ‘(+)’, namely ‘True’
      In the expression: 'x' + True
      In an equation for ‘x’: x = 'x' + True
(deferred type error)
*Optics.Core> let xs = view folded "foobar"

<interactive>:4:10: warning: [-Wdeferred-type-errors]
    • A_Fold cannot be used as A_Getter
      Perhaps you meant one of these:
        ‘traverseOf_’ (from Optics.Fold)
        ‘foldMapOf’ (from Optics.Fold)
        ‘toListOf’ (from Optics.Fold)
        ‘(^..)’ (from Optics.Operators)
    • In the expression: view folded "foobar"
      In an equation for ‘xs’: xs = view folded "foobar"
*Optics.Core> xs
*** Exception: unreachable
CallStack (from HasCallStack):
  error, called at src/Optics/Internal/Optic/Subtyping.hs:40:15 in optics-core-0.3-inplace:Optics.Internal.Optic.Subtyping
```

We reach: https://github.com/well-typed/optics/blob/a47f0047dc00b39cc31b02c3599cf4f419a6d7bc/optics-core/src/Optics/Internal/Optic/Subtyping.hs#L40

We need a support in GHC to scrutinize on empty dictionary, compare `EmptyCase` vs. `error "this should not happen"`.

In other words, We should be able to write something like 

```haskell
foo :: (Int ~ Char) => Void
foo ()
```

or

```haskell
bar :: TypeError ('Text "err") => Void
bar ()
```

[...]

PR #369, opened 2020-11-04
Simplify the error generated by the GeneralLabelOptic instance

arybczak:
Requested in feedback after the Haskell Exchange talk.

Before:
```
λ> data User = User { name :: String } deriving Show
λ> User "Tom" & #name .~ ("x")

<interactive>:2:1: error:
    • No instance for LabelOptic "name" ‘k’ ‘User’ ‘b’ ‘a’ ‘[Char]’
      Perhaps you forgot to define it or misspelled its name?
        it :: forall (k :: OpticKind) b a.
              ((TypeError ...), Is k A_Setter) =>
              b
```

After:
```
λ> data User = User { name :: String } deriving Show
λ> User "Tom" & #name .~ ("x")

<interactive>:2:14: error:
    • Type ‘User’ has no label ‘name’ as an optic for ‘a0’
    • In the first argument of ‘(.~)’, namely ‘#name’
      In the second argument of ‘(&)’, namely ‘#name .~ ("x")’
      In the expression: User "Tom" & #name .~ ("x")
```

### DIFF
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -217,15 +217,13 @@ instance {-# OVERLAPPABLE #-}
 -- variables and overlapping instances which are irrelevant and confusing. Use
 -- incoherent instance providing a custom type error to cut its efforts short.
 instance {-# INCOHERENT #-}
-  TypeError
-   ('Text "No instance for LabelOptic " ':<>: 'ShowType name
-    ':<>: 'Text " " ':<>: QuoteType k
-    ':<>: 'Text " " ':<>: QuoteType s
-    ':<>: 'Text " " ':<>: QuoteType t
-    ':<>: 'Text " " ':<>: QuoteType a
-    ':<>: 'Text " " ':<>: QuoteType b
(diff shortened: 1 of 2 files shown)


adamgundry:
The new message, and forcing `s` and `t` to have the same shape, look good to me.

We should really have doctests for type error messages... oh wait, that's #322.

arybczak:
> We should really have doctests for type error messages.

Indeed. #361 adds some, but there's still a lot missing.

arybczak:
Somewhat superseded by #361 as there the types are more refined.

```
λ> data User = User { name :: String } deriving Show
λ> User "Tom" & #name .~ ("x")

<interactive>:2:14: error:
    • No instance for LabelOptic "name" ‘A_Lens’ ‘User’ ‘User’ ‘String’ ‘[Char]’
      Possible solutions:
      - Check and correct spelling of the label
      - Define the LabelOptic instance by hand or via Template Haskell
      - Derive a Generic instance for ‘User’
    • In the first argument of ‘(.~)’, namely ‘#name’
      In the second argument of ‘(&)’, namely ‘#name .~ ("x")’
      In the expression: User "Tom" & #name .~ ("x")
```

So applying this patch loses some information.

PR #370, opened 2020-12-13
Use indexed-traversable

phadej:


arybczak:
Thanks.

Is there a plan to reverse dependencies of `inxeded-traversable-instances`?

phadej:
> Is there a plan to reverse dependencies of inxeded-traversable-instances?

From package description, https://hackage.haskell.org/package/indexed-traversable-instances-0.1/candidate

> The intention is to keep this package minimal; it provides instances that formely existed in lens or optics-extra. We recommend putting other instances directly into their defining packages. The `indexed-traversable` package is light, having only GHC boot libraries as its dependencies.

arybczak:
Right. I suppose we can reverse dependencies of at least `comonad` after.

phadej:
If you are worried about `comonad`, let me see if it's easy reverse now (without bumping oldest-supported GHC of `comonad`)

phadej:
Re `comonad`. See https://github.com/haskellari/indexed-traversable/pull/4#issuecomment-744023661

arybczak (approved):
Thanks :tada: Feel free to merge.

PR #373, opened 2020-12-15
Use elems for native toList in folds benchmark

phadej:


optics/benchmarks/folds.hs
```diff
@@ -167,7 +167,7 @@ main = defaultMainWith config
     ]
   , bgroup "hash map"
     [ bgroup "toList"
-      [ bench "native"          $ nf HM.keys h
+      [ bench "native"          $ nf HM.elems h
```
phadej, line 170: This was quite silly :)

arybczak (approved):

PR #377, opened 2020-12-24
Switch to GitHub Actions

arybczak:


arybczak:
@phadej  Why are the checks duplicated (pull_request & push)?

phadej:
- `pull_request` is when you target PR into that branch.
- `push` is when you push into the branch

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#on

phadej:
`haskell-ci` github actions backends lacks GHCJS support atm.

.github/workflows/haskell-ci.yml
```diff
@@ -0,0 +1,228 @@
+# This GitHub workflow config has been generated by a script via
+#
+#   haskell-ci 'github' 'cabal.project'
```
phadej, line 3: haskell-ci --config=cabal.haskell-ci cabal.project

arybczak:
@phadej Ah, I see. I found https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662/9 and applied a version of it to trigger builds only on pushes to master, but all PRs.

.github/workflows/haskell-ci.yml
arybczak: Oh. I see that also fixes the duplicate work problem :+1:

phadej (approved):
If this is unmodified workflow, then :+1;, let's switch.

I hope GHCJS support wont regress while I get to it on `haskell-ci` side.

PR #379, opened 2020-12-25
Port selected functions from Control.Lens.Plated

arybczak:
It needs a bit of polish wrt. documentation. Do we want to have it in 0.4?


optics-core/src/Optics/Plated.hs
```diff
@@ -0,0 +1,352 @@
...
+  -> (b -> m b) -> s -> m t
+transformMOnOf b l = traverseOf b . transformMOf l
+{-# INLINE transformMOnOf #-}
+
+-------------------------------------------------------------------------------
+-- Paramorphisms
+-------------------------------------------------------------------------------
+
+-- | Perform a fold-like computation on each value, technically a paramorphism.
+paraOf :: Is k A_Fold => Optic' k is a a -> (a -> [r] -> r) -> a -> r
```
michaelpj, line 321: I don't know to what degree we want to mimic the exact structure of `lens`, but arguably it would make sense to put the `*Of` variants in the modules for the optics they use, e.g. `paraOf` in `Fold`, `traverseOf` in `Setter` etc, leaving this module to contain the versions that actually use `Plated`.

(Which I just noticed is exactly how it is for `partsOf` already.)

optics-core/src/Optics/Plated.hs
```diff
@@ -0,0 +1,352 @@
...
+universe :: Plated a => a -> [a]
+universe = universeOf plate
+{-# INLINE universe #-}
+
+-- | Given a 'Fold' that knows how to locate immediate children, retrieve all of
+-- the transitive descendants of a node, including itself.
+universeOf :: Is k A_Fold => Optic' k is a a -> a -> [a]
+universeOf l = go
+  where
+    go a = a : foldMapOf l go a
```
michaelpj, line 184: I know `lens` doesn't do this either, but is there a reason not to define `universe*` fucntions in terms of `cosmos*` and `toListOf`?

michaelpj (commented):
I would definitely like to have these!

Boarders:
Any updates on this PR? I'd be happy to help get it over the line.

mikeplus64:
I wonder if `Plated` would be better off in its own package or in `recursion-schemes` where it IMO fits in very well. It's a really useful typeclass and it would be nice if `lens` and `optics` had compatibility with it as well.

```haskell
class Plated a where
  plate :: Applicative f => (a -> f b) -> s -> f t
```

I don't know if that's impossible with optics' `Traversal`.

optics-core/src/Optics/Plated.hs
```diff
@@ -0,0 +1,352 @@
...
+-- point.
+rewriteOf :: Is k A_Setter => Optic k is a b a b -> (b -> Maybe a) -> a -> b
+rewriteOf l f = go
+  where
+    go = transformOf l (\x -> maybe x go (f x))
+{-# INLINE rewriteOf #-}
+
+-- | Rewrite recursively over part of a larger structure.
+rewriteOn :: (Is k A_Setter, Plated a) => Optic k is s t a a -> (a -> Maybe a) -> s -> t
+rewriteOn b = over b . rewrite
```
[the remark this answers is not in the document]
arybczak: I copied the API of `lens` verbatim since I never used `Plated`, but you're right, it makes sense to provide MVP and extend if necessary.

optics-core/src/Optics/Plated.hs
arybczak: Indeed. I moved `paraOf` to `Optics.Fold` and removed `para` and `parts` for now.

optics-core/src/Optics/Plated.hs
arybczak: I have no idea, sadly I never used these functions so have no intuition behind them.

arybczak:
I've updated the PR and stripped the API down. Could you tell me if that's ok? I've never used `Plated` so can't really tell if the API is now much harder to use.

I'm even wondering about removing `Plate` and all functions without the `Of` suffix. But maybe that's too much?

optics-core/src/Optics/Plated.hs
arybczak: Oh, I didn't read this suggestion well enough it seems. Yeah, having `Optics.Plated` only have the functions with `plate` applied also sounds reasonable.

arybczak:
@mikeplus64 this is a good idea. `optics` can reuse `Plated` in the form you've written by simply wrapping it in `traversalVL`.

The only problem is the default instance would be missing, which might or might not be a big deal.

phadej:
In my opinion `Plated` type-class shouldn't be in `lens` and shouldn't be copied to `optics` either. It's just a traversal for immediate children. (It would be silly to have `Traversable` in `lens`, and we pulled `TraversableWithIndex` out of it too).

arybczak:
@phadej I'm with you on this one.

I think this PR can be merged as-is.

michaelpj (approved):
This looks great to me, very neat!

optics-core/src/Optics/Setter.hs
```diff
@@ -153,5 +155,29 @@ mapped :: Functor f => Setter (f a) (f b) a b
...
+-- point.
+rewriteOf :: Is k A_Setter => Optic k is a b a b -> (b -> Maybe a) -> a -> b
+rewriteOf l f = go
+  where
+    go = transformOf l (\x -> maybe x go (f x))
+{-# INLINE rewriteOf #-}
+
+-- | Transform every element by recursively applying a given 'Setter' in a
+-- bottom-up manner.
+transformOf :: Is k A_Setter => Optic k is a b a b -> (b -> b) -> a -> b
```
phadej, line 177: Is it deliberate  that the `*OnOf` variants are left out? They are useful when you e.g. want  to transform all `Pat`s inside `Exp`:

```haskell
transformOnOf patExps expPlated modifyPat :: Exp -> Exp where
  patExps :: Setter' Pat Exp
  expPlated :: Setter' Pat Pat
  modifyPlated :: Pat -> Pat
```

Of course I can do `over patExps (transformOf expPlated modifyPlated)`...

optics-core/src/Optics/Fold.hs
```diff
@@ -675,5 +678,29 @@ lookupOf :: (Is k A_Fold, Eq a) => Optic' k is s (a, v) -> a -> s -> Maybe v
 lookupOf o a = foldrOf o (\(a', v) next -> if a == a' then Just v else next) Nothing
 {-# INLINE lookupOf #-}
 
+-- | Given a 'Fold' that knows how to locate immediate children, retrieve all of
+-- the transitive descendants of a node, including itself.
+universeOf :: Is k A_Fold => Optic' k is a a -> a -> [a]
+universeOf l = go
+  where
+    go a = a : foldMapOf l go a
```
phadej, line 686: this is slow for trees. `++` can easily make this quadratic. I don't know why DList = Endo [a] is not used here instead.

optics-core/src/Optics/Setter.hs
arybczak: > Of course I can do over patExps (transformOf expPlated modifyPlated)...

Yeah, that's what was also mentioned in https://github.com/well-typed/optics/pull/379#discussion_r654792249, I think these are a bit excessive.

We can always add them later if there is enough demand.

optics-core/src/Optics/Fold.hs
arybczak: Can you give an example?

> I don't know why DList = Endo [a] is not used here (and in lens) instead.

Me neither, I assume since `lens` has it like that it's fine :thinking: It can be optimized later if necessary.

phadej: See https://github.com/ekmett/lens/pull/994

arybczak: I updated PR to reflect changes made to lens.

PR #381, opened 2020-12-28
Explicit imports

phadej:


arybczak:
Eh, I don't like explicit imports, they're really annoying when changing code :thinking:

phadej:
Yes, exactly. Their value is to make code be resilient to upstream changes.

If we want to use open imports, then our dependency bounds have to be `<A.B.C`.

arybczak:
Fair enough :+1: I suppose that code won't change much anyway.

PR #383, opened 2020-12-29
Fix doctests for GHC 8.2.2

arybczak:


optics-core/src/Optics/Generic.hs
```diff
@@ -82,7 +83,11 @@ data Void0
 --
 -- >>> user ^. gfield @"salary"
 -- ...
--- ...User’ doesn't have a field named ‘salary’
+#if __GLASGOW_HASKELL__ == 802
```
phadej, line 86: this actually may disable doctests all together (because an empty line may be introduced by cpphs, so we have just `...` which matches everything).

phadej: ... `cabal haddock` should tell that. (it might also depend on `cpp` used).

arybczak: I solved it properly (and improved error messages of generic optics in the process).

arybczak:
I think it's good now.

arybczak:
Btw, how does `doctest` work in the CI? If I try to invoke doctest commands locally, I just get errors about missing external imports.

```
unknown@electronics optics-core $ doctest  -XBangPatterns -XDefaultSignatures -XDeriveFunctor -XFlexibleContexts -XFlexibleInstances -XFunctionalDependencies -XGADTs -XInstanceSigs -XLambdaCase -XMultiParamTypeClasses -XRankNTypes -XScopedTypeVariables -XTupleSections -XTypeApplications -XTypeFamilies -XTypeOperators src

src/Optics/Internal/Bi.hs:12:1: error:
    Could not find module ‘Data.Profunctor.Indexed’
    Use -v to see a list of the files searched for.
   |
12 | import Data.Profunctor.Indexed
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

arybczak:
I've split parts into #384.

arybczak:
This doesn't work, `cabal haddock` cuts the documentation at the first `#if`, most likely because the preprocessor inserts line breaks.

PR #385, opened 2020-12-30
Temporary compatibility fix for GHC 9.0.1

arybczak:


optics/optics.cabal
```diff
@@ -168,6 +172,10 @@ benchmark folds
   if impl(ghcjs)
     buildable: False
 
+  -- Doesn't compile, pulls in Cabal 3.2.1.0 that fails
```
phadej, line 175: Update your `cabal-install` to `3.4` if you are testing GHC-9.0

phadej: https://mail.haskell.org/pipermail/cabal-devel/2020-October/010490.html

optics/optics.cabal
```diff
@@ -132,6 +132,10 @@ test-suite optics-tests
   if impl(ghcjs)
     buildable: False
 
+  -- Doesn't compile because of inspection-testing
```
phadej, line 135: There is patch in [head.hackage](https://gitlab.haskell.org/ghc/head.hackage), https://gitlab.haskell.org/ghc/head.hackage/-/blob/master/patches/inspection-testing-0.4.2.4.patch,

doesn't it work?

arybczak:
Thanks, I forgot about head.hackage. Both tests and benchmarks work with cabal-install 3.4.0.0 and head.hackage.

PR #388, opened 2021-01-04
Make selfIndex :: IxIso

phadej:


arybczak:
AFAIR we have `selfIndex` as `IxGetter` and not as e.g. `IxLens` to not make `indices` unsafe :thinking:

phadej:
Yes. Maybe this new thing should rather be `unsafeSelfIndex`, as `selfIndex` with `indices` allow to implement what we call `unsafeFiltered`.

I'm fine with that.

The questions in code are still relevant:
- Were this (`unsafeSelfIndex`) should be. Should we have IxIso?

arybczak:
> Should we have IxIso?

`IxIso` sounds a bit weird to me. `Iso' s a` means that `s` and `a` are somewhat equivalent, so what additional information would you want to attach as an index?

phadej:
> > Should we have IxIso?
> 
> `IxIso` sounds a bit weird to me. `Iso' s a` means that `s` and `a` are somewhat equivalent, so what additional information would you want to attach as an index?

Any. `IxIso' i s a ~ (s -> (i, a), a -> s)`, so indeed index is something derived from `s` / `a`.

My example

```haskell
-- >>> iover (mapped % selfIndex % mapped) (\xs x -> x + maximum xs) [[1,2,3],[4,5,6],[7,8,9]]
-- [[4,5,6],[10,11,12],[16,17,18]]
```

tries to illustate where it could be useful. To preserve some "mid element" even when there is further focusing.

---

Before I forget, `IxLens i (i, a) (i, b) a b` would be useful too. Not sure if `lens` has a name for it.

phadej:
FWIW, `lens` `selfIndex` behaves unsafely (which probably isn't news):

```haskell
*Control.Lens> iover (mapped .> selfIndex <. mapped) (\xs x -> x + maximum xs) [[1,2,3],[4,5,6],[7,8,9]]
[[4,5,6],[10,11,12],[16,17,18]]
```

arybczak:
IMO if you want to have `unsafeSelfIndex`, we should just make it `IxLens`. I'm not convinced this warrants changing `ixcontramap`.

phadej:
Yes, I agree `IxLens` should be enough for all examples I can think about.

I also wrote down some thoughts related `IxLens`: https://oleg.fi/gists/posts/2021-01-08-indexed-optics-dilemma.html

arybczak:
Shall we close this?

PR #389, opened 2021-01-06
Add ifst and isnd

phadej:


phadej:
The name focus comes from

```haskell
-- (Actually, AffineTraversable, and Traversable1 would be more right superclasses, if also existed).
class Traversable t => Singleton t where
    focus :: Functor f => (a -> f b) -> t a -> f (t b)
```

A name I invented just today, so I'm not (yet) attached to it.

https://oleg.fi/gists/posts/2018-05-12-singleton-container.html uses `peekaboo` :)

arybczak:
`ifocus` is a very generic name :thinking: I take it `lens` doesn't have this?

phadej:
`ifst` and `isnd` would make sense. (`first` is slightly too overloaded). Also they are about as short as `_1` and `_2`.

arybczak:
Sounds good :+1:

arybczak (approved):

phadej:
@adamgundry are you fine with this?

adamgundry (approved):
Seems fine to me. Sorry for the latency.

phadej:
@adamgundry no worries. As long as you find time to discuss *breaking changes*, there is no problem delaying things like this. :)

PR #390, opened 2021-01-10
Bump version to 0.4, add preliminary and incomplete changelog for opt…

phadej:
…ics-core

I need a version bump, so I can prepare for a new release. (Getting CPP right).

I added some changelog, what I could tell from API diff. There are few `@since` annotations. (Probably too few).

I think @arybczak is the best one to write about changes related to `LabelOptic`

### DIFF
--- a/template-haskell-optics/CHANGELOG.md
+++ b/template-haskell-optics/CHANGELOG.md
@@ -1,2 +1,5 @@
+# template-haskell-optics-0.2
+* TBW: No DataPrism
+
 # template-haskell-optics-0.1 (2020-08-29)
 * Initial release
--- a/migration-guide-0.4.md
+++ b/migration-guide-0.4.md
@@ -1,0 +1,42 @@
+# Migration guide to optics-0.4
+
+## FunctorWithIndex instances
+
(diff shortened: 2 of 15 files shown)


arybczak:
I think migration guide should be separate from the changelog.

phadej:
If you want to separate it, feel free. Remember, Hackage shows only Readme and a link to changelog, so if we separate migration guide out of changelog, we should have a link there.

phadej:
I have split the migration notes into separate file.

template-haskell-optics/CHANGELOG.md
```diff
@@ -1,2 +1,5 @@
+# template-haskell-optics-0.2
+* TBW: No DataPrism
```
arybczak, line 2: `DataPrism` wasn't exported, so it probably doesn't matter.

phadej: I defer completion and correction of that changelog to you for later.

phadej:
I think this is good enough. Further improvements can be done in the follow-ups.

PR #391, opened 2021-01-16
Compare functions from Optics.Generic with generic-optics

arybczak:


arybczak:
@adamgundry  would you mind having a quick look?

adamgundry:
@arybczak thanks for writing this up. I agree it makes sense to have the `Optics.Generic` haddocks contain a pointer to `generic-optics` and a brief comparison. It doesn't need to be defensive about the fact there is duplication (Hackage is full of that!). Rather let's try to explain the trade-offs for the user. We could point out that `generic-optics` has functionality not included in `Optics.Generic`, because the latter deliberately includes a smaller API.

I'm less sure about putting the detailed performance comparison in the haddocks, since things may change in the future (with new GHC releases, even holding both library versions fixed) and we may want to refine the methodology. What do you think about moving the Performance section to a file in this repo and merely linking to it from the haddocks?

IMHO the tests in the "User experience" section don't need to show up in the generated haddocks. The doctests are good to have, but perhaps we could put them in an unused named chunk so that haddock drops them?

arybczak:
I'm inclined to drop this one.

PR #393, opened 2021-01-18
Add failing for traversals

arybczak:
We have `adjoin`, might as well add `failing`.

The awkwardness here is about naming, since this PR renames existing `Fold` equivalents.

There is also an issue that `afailing` for traversals is safe, while `failing` for traversals is safe only on disjoint traversals.

Perhaps we want a different name than `failing` to avoid a minor API breakage (similar to how we have `summing` for folds and `adjoin` for traversals). I have no idea what is a good name though.

### DIFF
--- a/optics-core/src/Optics/AffineFold.hs
+++ b/optics-core/src/Optics/AffineFold.hs
@@ -46,7 +46,7 @@ module Optics.AffineFold
   -- there is not a unique choice of monoid to use that works for all optics,
   -- and the ('<>') operator could not be used to combine optics of different
   -- kinds.
-  , afailing
+  , afailing_
 
   -- * Subtyping
   , An_AffineFold
@@ -127,20 +127,20 @@ filtered p = afoldVL (\point f a -> if p a then f a else point a)
 
 -- | Try the first 'AffineFold'. If it returns no entry, try the second one.
 --
--- >>> preview (ix 1 % re _Left `afailing` ix 2 % re _Right) [0,1,2,3]
+-- >>> preview (ix 1 % re _Left `afailing_` ix 2 % re _Right) [0,1,2,3]
 -- Just (Left 1)
 --
--- >>> preview (ix 42 % re _Left `afailing` ix 2 % re _Right) [0,1,2,3]
+-- >>> preview (ix 42 % re _Left `afailing_` ix 2 % re _Right) [0,1,2,3]
 -- Just (Right 2)
 --
-afailing
+afailing_
   :: (Is k An_AffineFold, Is l An_AffineFold)
   => Optic' k is s a
   -> Optic' l js s a
   -> AffineFold s a
-afailing a b = afolding $ \s -> maybe (preview b s) Just (preview a s)
-infixl 3 `afailing` -- Same as (<|>)
-{-# INLINE afailing #-}
+afailing_ a b = afolding $ \s -> maybe (preview b s) Just (preview a s)
+infixl 3 `afailing_` -- Same as (<|>)
+{-# INLINE afailing_ #-}
 
 -- | Check to see if this 'AffineFold' doesn't match.
 --
--- a/optics-core/src/Optics/AffineTraversal.hs
+++ b/optics-core/src/Optics/AffineTraversal.hs
@@ -48,6 +48,8 @@ module Optics.AffineTraversal
   -- * Additional elimination forms
   , withAffineTraversal
 
+  , afailing
+
   -- * Subtyping
   , An_AffineTraversal
   -- | <<diagrams/AffineTraversal.png AffineTraversal in the optics hierarchy>>
@@ -63,6 +65,7 @@ module Optics.AffineTraversal
 import Data.Profunctor.Indexed
 
 import Optics.Internal.Optic
+import Optics.Internal.Utils
 
 -- | Type synonym for a type-modifying affine traversal.
 type AffineTraversal s t a b = Optic An_AffineTraversal NoIx s t a b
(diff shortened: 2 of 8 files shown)


arybczak:
@phadej @adamgundry  thoughts?

georgefst:
I just spent quite a while looking for this, while porting some `lens` code that just used `(<>)` with `AffineTraversal`s (in the end, the existing `afailing` turned out to be just about good enough for my use case, for now).

I'd suggest that this PR should also add a `Monoid structure` section to the docs for `Traversal`, like [`Fold`](https://hackage.haskell.org/package/optics-core-0.4/docs/Optics-Fold.html#g:monoids), [`AffineFold`](https://hackage.haskell.org/package/optics-core-0.4/docs/Optics-AffineFold.html#g:7) and [`Traversal`](https://hackage.haskell.org/package/optics-core-0.4/docs/Optics-Traversal.html#g:9) have. PS. there are minor inconsistencies in those three existing sections in the title and URL.

cdfa:
I would like to see `failing` variations for other optics too! However, instead of coming up with clever name variations, maybe we can generalize is into one `gfailing` (and `igfailing` for indexed optics).
Additionally, it would be nice if the kind of the "recovery" optic would be preserved in some cases. For example, combining in affine fold with a getter could yield a getter.
I had a go at this and you can find the result below.

Additionally, I think we could refine the precondition that the combined optics are disjoint a little more to allow for things like ``ix i \`gfailing\` _last``.
The intuition is that ``k \`gfailing\` l`` should be legal when the success/failure of `k` is independent of the values of `l`'s foci.
More accurately stated: ``k \`gfailing\` l`` is legal when `k` and `l` are disjoint *or* `l` focuses at least as "deep" as `k`.

<details>
<summary>Optics.Fallible</summary>

```
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE FunctionalDependencies #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TypeApplications #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeOperators #-}

module Optics.Fallible where

import Data.Type.Equality

import Optics
import Optics.Internal.Utils

class FallibleOptic k l m s t a b | k l -> m where
    {-|
    NB: @gfailing k l@ only produces legal R/W optics when the foci of @k@ and @l@ are disjoint or @l@ focuses at least as "deep" as @k@.
    For example, @_tail \`gfailing\` simple@ is illegal, because composition is not preserved, i.e. with @o = _tail \`gfailing\` simple@, @Identity . cons 1@ and @g = Identity . cons 2@, we have
    
    @
    getCompose $ traverseOf o (Compose . fmap f . g) [] ≡ Identity (Identity [ 1 , 2 ])
        /= Identity (Identity [ 2 , 1 ]) ≡ fmap (traverseOf o f) $ traverseOf o g []
    @

    On the other hand, @ix i \`gfailing\` _last@ would be legal.
    In other words, the success/failure of @k@ cannot depend on the values of @l@'s foci.
    -}
    gfailing
        :: Optic k is s t a b -> Optic l is s t a b -> Optic m NoIx s t a b

infixl 3 `gfailing` -- Same as (<|>)

-- Prism
instance FallibleOptic A_Prism An_Iso A_Lens s t a b where
    gfailing k l = gfailing k $ castOptic @A_Lens l

instance FallibleOptic A_Prism A_Lens A_Lens s t a b where
    gfailing k l = gfailing (castOptic @An_AffineTraversal k) l

instance FallibleOptic A_Prism A_Prism An_AffineTraversal s t a b where
    gfailing = gfailing `on` castOptic @An_AffineTraversal

instance FallibleOptic A_Prism A_Getter A_Getter s s a a where
    gfailing k l = gfailing (castOptic @An_AffineFold k) l

-- AffineTraversal
instance FallibleOptic An_AffineTraversal A_Lens A_Lens s t a b where
    gfailing k l = withAffineTraversal k $ \matchK _ -> withLens l $ \viewL _ ->
        lens (\s -> fromRight (viewL s) $ matchK s)
             (\s b -> fromMaybe (set l b s) $ failover k (const b) s)

instance FallibleOptic An_AffineTraversal An_AffineTraversal An_AffineTraversal s t a b where
    gfailing k l = atraversalVL $ \point f s -> let
        OrT visited fu = atraverseOf k (OrT False . point) (wrapOrT . f) s
        in if visited then fu else atraverseOf l point f s

-- AffineFold
instance FallibleOptic An_AffineFold A_Getter A_Getter s s a a where
    gfailing k l = to $ \s -> fromMaybe (view l s) $ preview k s

instance FallibleOptic An_AffineFold An_AffineFold An_AffineFold s s a a where
    gfailing k l = afolding $ \s -> preview l s <|> preview k s

-- Traversal
instance FallibleOptic A_Traversal A_Traversal A_Traversal s t a b where
    gfailing k l = traversalVL $ \f s -> let
        OrT visited fu = traverseOf k (wrapOrT . f) s
        in if visited then fu else traverseOf l f s

-- Fold
instance FallibleOptic A_Fold A_Fold A_Fold s s a a where
    gfailing k l = foldVL $ \f s -> let
        OrT visited fu = traverseOf_ k (wrapOrT . f) s
        in if visited then fu else traverseOf_ l f s

instance {-# OVERLAPPABLE #-}( JoinKinds k l m
                             , Is k m
                             , Is l m
                             , FallibleOptic m m m s t a b
                             , (k == l) ~ 'False
                             ) => FallibleOptic k l m s t a b where
    gfailing k l = gfailing (castOptic @m k) (castOptic @m l)
```

</details>

I can make a full PR if you're interested.

Lev135:
> There is also an issue that afailing for traversals is safe, while failing for traversals is safe only on disjoint traversals.

Maybe it should be named `unsafeAfailing` like `unsafeFiltered`?

ozkutuk:
I would also like to have this `optics`, specifically `afailing` (as it is currently in the PR) for combining a disjoint pair of `AffineTraversal`s. I am currently using `aadjoin o1 o2 = singular (adjoin o1 o2)` as a substitute, but it is less than ideal.

adamgundry:
Apologies, I hadn't really taken this in before now. Some scattered thoughts:

* If I understand correctly, (new) `failing` for traversals is to (new) `failing_` for folds as `adjoin` is to `summing`. That, as well as the difference in side conditions, makes me think a new name might be no bad thing. Otherwise the question arises as to why we don't have `summing`/`summing_` instead of `adjoin`/`summing`.
* Perhaps `disjoin` for traversals is a possible new name (then we could keep `failing` for folds)?
* The situation with `afailing` for affine traversals is slightly different in that it doesn't have the side condition, and of course we don't have an `adjoin`/`summing` equivalent as a guide. Is `adisjoin` plausible or is that too ugly? (If we were to keep `failing` for folds, it would seem odd if `afailing` applied to affine traversals rather than affine folds.)
* I'm not fundamentally opposed to making a breaking change in a corner like this, as I think the eventual consistency of the API is the priority. But it's not obvious to me that changing `failing` is actually better for API consistency.
* In any case, we should definitely explain the relationship between the various versions in the Haddocks.
* Also, the PR seems to be missing a replacement `iafailing`?

arybczak:
FYI I don't have steam at the moment to get this over the finish line. If someone wants to overtake the PR and adjust it to be in a mergable state, be my guest :+1:

Issue #137, opened 2019-02-24
Experiment with less INLINE pragmas


arybczak:
We currently mark everything as INLINE. Some of these are needed for rewrite rules or optimizing away the profunctor typeclasses, but excessive use of these is known to slow down compilation and increase code size.

Case in point: https://mpickering.github.io/posts/2017-05-17-inlining-case-study.html

We should probably figure out which functions need to be marked as such and which aren't. We have reliable way of testing that thanks to `inspection-testing`.

arybczak:
Preliminary testing shows that functions that consume optics and optics itself usually (unless they're really small) need to be marked as INLINE for the optimization to happen.

Looks like implementations of profunctor instances and other utilities that sit deeper don't need INLINE pragmas though.

PR #394, opened 2021-01-31
Reduce the amount of INLINE pragmas in indexed-profunctors and optics-core

arybczak:
Part 1/x of #137.


arybczak:
This doesn't affect performance, but seems to slightly reduce compilation times.

arybczak:
I'll merge after the weekend unless anyone objects.

adamgundry:
Thanks for looking at this. :+1:

PR #397, opened 2021-02-04
Remove unsafeCoerce from appendIndices and fix tests with GHC 9.0.1

### LINKED ISSUE #396: Tests fail with 9.0.1 because of unsafeEqualityProof
15 of tests fail with GHC 9.0.1, mostly because of unsafeEqualityProof, e.g.

```haskell
        tests/Optics/Tests/Utils.hs:56:
        tests/Optics/Tests/Labels/TH.hs:61:20: label5lhs ==- label5rhs failed:            
            LHS:                                                                          
                label5lhs = label5lhs_szze `cast` <Co:31>                                 
                                                                                          
                label5lhs_szze                                                            
                  = \ @a @b s name_ age_ fishName_ pets_ ->                               
                      case unsafeEqualityProof of { UnsafeRefl co2 ->
                      case s of { Human x1_af3x x2_af3y x3_af3z x4_af3A ->
                      (Human
                         name_
                         age_
                         (case x3_af3z of {
                            GoldFish x1_aeS2 -> GoldFish fishName_;
                            Herring x1_aeS4 -> Herring fishName_
                          })
                         pets_)
                      `cast` <Co:4>
                      }
                      }
                
            RHS:
                label5rhs
                  = \ @a @b s name_ age_ fishName_ pets_ ->
                      case s of { Human ds_dpF7 ds_dpF8 ds_dpF9 ds_dpFa ->
                      Human
                        name_
                        age_
                        (case ds_dpF9 of {
                           GoldFish ds_dpEY -> GoldFish fishName_;
                           Herring ds_dpEZ -> Herring fishName_
                         })
                        pets_
                      }
```

Does anyone know why are these in a release? I know they used to show in GHC compiled from git in the past, but they didn't in releases.

arybczak:
Fixes #396.

This doesn't slow compilation, I tested with a big project.

### DIFF
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -30,7 +30,6 @@ module Optics.Internal.Optic
   ) where
 
 import Data.Function ((&))
-import Data.Type.Equality
 
 import Data.Profunctor.Indexed
 
@@ -114,7 +113,7 @@ castOptic (Optic o) = Optic (cast o)
 -- indexed, the composition preserves all the indices.
 --
 infixl 9 %
-(%) :: (Is k m, Is l m, m ~ Join k l, ks ~ Append is js)
+(%) :: (Is k m, Is l m, m ~ Join k l, AppendIndices is js, ks ~ Append is js)
     => Optic k is s t u v
     -> Optic l js u v a b
     -> Optic m ks s t a b
@@ -127,14 +126,14 @@ o % o' = castOptic o %% castOptic o'
 -- type inference if the type of one of the optics is otherwise
 -- under-constrained.
 infixl 9 %%
-(%%) :: forall k is js ks s t u v a b. ks ~ Append is js
+(%%) :: forall k is js ks s t u v a b. (AppendIndices is js, ks ~ Append is js)
      => Optic k is s t u v
      -> Optic k js u v a b
      -> Optic k ks s t a b
 Optic o %% Optic o' = Optic oo
   where
     oo :: forall p i. (Profunctor p, Constraints k p) => Optic__ p i (Curry ks i) s t a b
-    oo | Refl <- appendIndices @is @js @i = o . o'
+    oo | IxEq <- appendIndices @is @js @i = o . o'
 
 -- | Flipped function application, specialised to optics and binding tightly.
 --
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -113,7 +113,7 @@ castOptic (Optic o) = Optic (cast o)
 -- indexed, the composition preserves all the indices.
 --
 infixl 9 %
-(%) :: (Is k m, Is l m, m ~ Join k l, AppendIndices is js, ks ~ Append is js)
+(%) :: (Is k m, Is l m, m ~ Join k l, AppendIndices is js ks)
     => Optic k is s t u v
     -> Optic l js u v a b
     -> Optic m ks s t a b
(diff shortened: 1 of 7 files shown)


phadej:
Let's see after `INLINE`s are removed.

arybczak:
I also tried that with the inlineless branch and it was fine 👍

optics-core/src/Optics/Internal/Optic/TypeLevel.hs
```diff
@@ -98,37 +96,34 @@ instance CurryCompose xs => CurryCompose (x ': xs) where
 ----------------------------------------
 -- Indices
 
+-- | Tagged version of 'Data.Type.Equality.(:~:)' for carrying evidence that two
+-- index lists in a curried form are equal.
+data IxEq i is js where
```
phadej, line 101: This is nice.

optics-core/src/Optics/Internal/Optic.hs
```diff
@@ -115,7 +114,7 @@ castOptic (Optic o) = Optic (cast o)
 -- indexed, the composition preserves all the indices.
 --
 infixl 9 %
-(%) :: (Is k m, Is l m, m ~ Join k l, ks ~ Append is js)
+(%) :: (Is k m, Is l m, m ~ Join k l, AppendIndices is js, ks ~ Append is js)
```
adamgundry, line 117: I think this is generally a good idea, but the only thing that is a slight shame is making the type of `%` more complicated, with `AppendIndices` potentially showing up in inferred types. Perhaps it is worth making `AppendIndices` takes `ks` as a parameter so that `AppendIndices is js ks` can imply `ks ~ Append is js`?

phadej: One could maybe get rid of `Append` family all together, by having 

```diff
-appendIndices :: IxEq i (Curry xs (Curry ys i)) (Curry (Append xs ys) i)
+appendIndices :: IxEq i (Curry xs (Curry ys i)) (Curry zs i)
```

... but I guess we cannot as there is no sane way to break functional dependencies in GHCs.

arybczak: > Perhaps it is worth making AppendIndices takes ks as a parameter

That's what I've done in #353, but then you pointed out [here](https://github.com/well-typed/optics/pull/353#issuecomment-687915488) that GHC expands `ks` to be `Append is js` and removing it would be better ;)

optics-core/src/Optics/Internal/Optic/TypeLevel.hs
```diff
@@ -107,23 +107,20 @@ data IxEq i is js where
...
 --    where f = (->)
 -- @
-class AppendIndices xs ys where
-  appendIndices :: IxEq i (Curry xs (Curry ys i)) (Curry (Append xs ys) i)
+class AppendIndices xs ys ks | xs ys -> ks where
+  appendIndices :: IxEq i (Curry xs (Curry ys i)) (Curry ks i)
 
--- | If we know the second list is empty, we can pick the first list without
--- knowing anything about it, hence the instance is marked as INCOHERENT.
-instance {-# INCOHERENT #-} AppendIndices xs '[] where
```
phadej, line 112: Oh, I thought that `FunctionalDependencies` would choke on this? Is `INCOHERENT` lifting that somehow?

optics-core/src/Optics/Internal/Optic.hs
arybczak: > One could maybe get rid of Append family all together

I tried that and looks like it works, so I re-added the third parameter and removed dependence on `Append`.

optics-core/src/Optics/Internal/Optic/TypeLevel.hs
```diff
@@ -98,37 +96,31 @@ instance CurryCompose xs => CurryCompose (x ': xs) where
...
+-- immediately.
+instance {-# INCOHERENT #-} AppendIndices xs '[] xs where
+  appendIndices = IxEq
+
+instance AppendIndices '[] ys ys where
+  appendIndices = IxEq
+
+instance AppendIndices xs ys ks => AppendIndices (x ': xs) ys (x ': ks) where
+  appendIndices :: forall i. IxEq i (Curry (x ': xs) (Curry ys i)) (Curry (x ': ks) i)
+  appendIndices | IxEq <- appendIndices @xs @ys @ks @i = IxEq
```
phadej, line 123: This looks very nice. :+1:

arybczak:
It's worth noting that due to `%` being `infixl` when composing multiple indexed optics the behaviour of AppendIndices is quadratic :thinking: 

Though that shouldn't be a problem in practice since I doubt anyone uses >5 indices in a single pipeline.

phadej:
re quadratic: wasn't that the case with `Append` also?

arybczak:
Oh, you're right. I thought `unsafeCoerce` saved us there before, but indeed `Append` was still quadratic.

optics-core/src/Optics/Internal/Optic/TypeLevel.hs
arybczak: It's not INCOHERENT, it's just that the consistency check for fundeps is quite lenient.

Issue #368, opened 2020-10-29
Composition fails with overlapping instances


dminuoso:
Hi, after parameterizing a function by an arbitrary setter, I've stumbled over this seemingly strange error:

```
g :: Lens' String Char                                                                                   
g = undefined                                                                                            
                                                                                                         
f :: Is k A_Setter => Optic' k is Char Int -> String                                                     
f o = "foo" & (g%o) .~ 10     
```

```
tool/Parser.hs:76:15: error:
    • Overlapping instances for Is (Join A_Lens k) A_Setter
        arising from a use of ‘.~’
      Matching instances:
        instance [overlappable] (TypeError ...) => Is k l
          -- Defined in ‘optics-core-0.3.0.1:Optics.Internal.Optic.Subtyping’
        instance Is k k
          -- Defined in ‘optics-core-0.3.0.1:Optics.Internal.Optic.Subtyping’
        instance Is A_Lens A_Setter
          -- Defined in ‘optics-core-0.3.0.1:Optics.Internal.Optic.Subtyping’
        ...plus four others
        (use -fprint-potential-instances to see them all)
      (The choice depends on the instantiation of ‘k’
       To pick the first instance above, use IncoherentInstances
       when compiling the other instance declarations)
    • In the second argument of ‘(&)’, namely ‘(g % o) .~ 10’
      In the expression: "foo" & (g % o) .~ 10
      In an equation for ‘f’: f o = "foo" & (g % o) .~ 10
   |
76 | f o = "foo" & (g%o) .~ 10
   |               ^^^^^^^^^^^

tool/Parser.hs:76:16: error:
    • Overlapping instances for Is k (Join A_Lens k)
        arising from a use of ‘%’
      Matching instances:
        instance [overlappable] (TypeError ...) => Is k l
          -- Defined in ‘optics-core-0.3.0.1:Optics.Internal.Optic.Subtyping’
        instance Is k k
          -- Defined in ‘optics-core-0.3.0.1:Optics.Internal.Optic.Subtyping’
        instance Is A_Getter A_Fold
          -- Defined in ‘optics-core-0.3.0.1:Optics.Internal.Optic.Subtyping’
        ...plus 35 others
        (use -fprint-potential-instances to see them all)
      (The choice depends on the instantiation of ‘k’
       To pick the first instance above, use IncoherentInstances
       when compiling the other instance declarations)
    • In the first argument of ‘(.~)’, namely ‘(g % o)’
[...]

PR #399, opened 2021-02-08
Turn Join into a class for neater type signatures

arybczak:
Somewhat related to https://gitlab.haskell.org/ghc/ghc/-/issues/19336 (it's a GHC bug, but it pushed me to find a solution for using a class for `Join` that works).

I wanted to do that for some time now, considering that the type signature of `%` was a bit messy, not to mention type signatures of generic optic functions (see e.g. #368). Now looks like a good time to do that since `Append` was also changed.

Now the signature of `%` is quite neat:

```haskell
>>> :t (%)
(%)
  :: (JoinKinds k l m, AppendIndices is js ks) =>
     Optic k is s t u v -> Optic l js u v a b -> Optic m ks s t a b
```

Also, now label compositions are somewhat readable:

```haskell
>>> :t #pets % traversed % #_GoldFish % #name
#pets % traversed % #_GoldFish % #name
  :: (Traversable t1, JoinKinds k1 l1 m, JoinKinds k2 l2 k1,
      JoinKinds k3 A_Traversal k2, LabelOptic "_GoldFish" l2 u1 v1 u2 v2,
      LabelOptic "name" l1 u2 v2 a b,
      LabelOptic "pets" k3 s t2 (t1 u1) (t1 v1)) =>
     Optic m NoIx s t2 a b
```

instead of this monstrosity

```haskell
>>> :t #pets % traversed % #_GoldFish % #name
#pets % traversed % #_GoldFish % #name
  :: (Traversable t1,
      Is
        (Join (Join k A_Traversal) l1)
        (Join (Join (Join k A_Traversal) l1) l2),
      Is l2 (Join (Join (Join k A_Traversal) l1) l2),
      Is (Join k A_Traversal) (Join (Join k A_Traversal) l1),
      Is l1 (Join (Join k A_Traversal) l1), Is k (Join k A_Traversal),
      Is A_Traversal (Join k A_Traversal),
      LabelOptic "_GoldFish" l1 u1 v1 u2 v2,
      LabelOptic "name" l2 u2 v2 a b,
      LabelOptic "pets" k s t2 (t1 u1) (t1 v1)) =>
     Optic (Join (Join (Join k A_Traversal) l1) l2) NoIx s t2 a b
```

See also change to `label6setter` in code, no more weird `Is` constraints.

If no one protests, I'll adjust codegen.

### DIFF
--- a/optics-core/src/Optics/Internal/Optic.hs
+++ b/optics-core/src/Optics/Internal/Optic.hs
@@ -113,12 +113,21 @@ castOptic (Optic o) = Optic (cast o)
 -- indexed, the composition preserves all the indices.
 --
 infixl 9 %
-(%) :: (Is k m, Is l m, m ~ Join k l, AppendIndices is js ks)
+(%) :: forall k l m is js ks s t u v a b. (JoinKinds k l m, AppendIndices is js ks)
     => Optic k is s t u v
     -> Optic l js u v a b
     -> Optic m ks s t a b
-o % o' = castOptic o %% castOptic o'
-{-# INLINE (%) #-}
+Optic k % Optic l = Optic m
+  where
+    km :: forall p i. Profunctor p => Optic_ m p i (Curry is i) s t u v
+    km = joinKinds @k @l @m @p k
+
+    lm :: forall p i. Profunctor p => Optic_ m p i (Curry js i) u v a b
+    lm = joinKinds @k @l @m @p l
+
+    m :: forall p i. (Profunctor p, Constraints m p)
+      => Optic__ p i (Curry ks i) s t a b
+    m | IxEq <- appendIndices @is @js @ks @i = km . lm
 
 -- | Compose two optics of the same flavour.
 --
@@ -132,7 +141,8 @@ infixl 9 %%
      -> Optic k ks s t a b
 Optic o %% Optic o' = Optic oo
   where
-    oo :: forall p i. (Profunctor p, Constraints k p) => Optic__ p i (Curry ks i) s t a b
+    oo :: forall p i. (Profunctor p, Constraints k p)
+       => Optic__ p i (Curry ks i) s t a b
     oo | IxEq <- appendIndices @is @js @ks @i = o . o'
 
 -- | Flipped function application, specialised to optics and binding tightly.
--- a/codegen/Subtypes.hs
+++ b/codegen/Subtypes.hs
@@ -97,7 +97,9 @@ genJoin = either (fail . show) id $ runG opticsKind $ \g -> do
     for_ gVertices $ \a -> do
         let k = gFromVertex a
         putStrLn   ""
-        putStrLn $ "  -- " ++ show k ++ "-----"
+        putStrLn $ "-- " ++ show k ++ " -----"
+
+        putStrLn $ joinInstance k k k
 
         for_ gVertices $ \b -> unless (a == b) $ do
             let l = gFromVertex b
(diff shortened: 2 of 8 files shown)


optics-core/src/Optics/Internal/Optic.hs
```diff
@@ -113,12 +113,21 @@ castOptic (Optic o) = Optic (cast o)
...
     => Optic k is s t u v
     -> Optic l js u v a b
-    -> Optic m ks s t a b
-o % o' = castOptic o %% castOptic o'
-{-# INLINE (%) #-}
+      -> Optic m ks s t a b
+Optic k % Optic l = Optic m
+  where
+    km :: forall p i. Profunctor p => Optic_ m p i (Curry is i) s t u v
+    km = joinKinds @k @l @m @p k
```
phadej, line 123: These `joinKinds` are essentially the same as `castOptic`? Should we have `castOptic`-like function which uses `joinKinds`?

Do we need `Is` at all, isn't `Is k l` `JoinKinds k l l`? I don't suggest we need to remove `Is`, I just wonder how `JoinKinds` is related.

arybczak: > These joinKinds are essentially the same as castOptic?

Yes, it's basically an extended version of `implies` you can use for both `k` and `l`.

> Should we have castOptic-like function which uses joinKinds?

`Is` is used for subtyping, while `JoinKinds` for composition. It's true that they look very similar, but have distinct use cases.

> Do we need Is at all, isn't Is k l JoinKinds k l l?

Yes, it looks like it. Functionally it could be used in place of `Is`, but it would be weird to users. Plus error messages in overlappable instances are different.

phadej: I agree that their use cases are different, and thus repeating is not bad (especially as it is codegen).

arybczak:
BTW I noticed that type signatures before were actually quadratic in terms of length of the pipeline, since e.g. in a signature for `#pets % traversed % #_GoldFish % #name` of length 4 you can see 3 nested joins, 2 nested joins and 1 nested join (multiple times actually).

The typeclass is linear.

optics-core/src/Optics/Internal/Optic.hs
adamgundry: Is there a reason to prefer this encoding over reusing `Is` like this?
```hs
class (Is k m, Is l m) => JoinKinds k l m | k l -> m
```

adamgundry (commented):
This generally looks good to me. I guess we are losing a small amount of expressiveness by dropping the type family, e.g. you needed to (re)introduce `ReadOnlyOptic`, but I'm inclined to think the more explicit approach is better. The type inference improvements are very welcome!

optics-core/src/Optics/Internal/Optic.hs
phadej: `JoinKinds` won't be `newtype` anymore. It might affect optimizer. If it doesn't, I'd like this sanity check.

arybczak: > Is there a reason to prefer this encoding over reusing `Is` like this?
> 
> ```haskell
> class (Is k m, Is l m) => JoinKinds k l m | k l -> m
> ```

We'd need to enumerate all cases anyway and we use the same codegen for it as for the `Is` instances.

I also don't know if there can be any complications like errors from `Is` being printed when overlappable error instance for `JoinKinds` is picked.

arybczak:
I'll merge after the weekend if there are no objections.

optics-core/src/Optics/Internal/Optic.hs
adamgundry: Fair points. I'm happy with the plan here, feel free to merge.

I can imagine one might know `JoinKinds` and yet need `Is`, but only if one is doing something pretty close to `optics` internals, so let's worry about that if we ever find such a case.

In another context (`HasField`-related) I find myself wishing that one could define magically-contracted constraint synonyms, e.g. define
```hs
type Is k l = JoinKinds k l l
```
and have GHC print `Is` instead of `JoinKinds` in inferred types and error messages whenever they are equivalent. Perhaps one day...

Issue #375, opened 2020-12-23
optics-0.4 release


phadej:
I'm very eager to use `indexed-traversable`.

I haven't followed the `LabelOptic` changes, is the #365 only things left?

Does @adamgundry or @arybczak have some breaking changes in mind, which would be great to get in?

I think it would be great to have GHC-9.0 compatible release as soon as it is released, so that could be a (soft) deadline for `optics-0.4`.

arybczak:
Apart from #365 I need to port a bit from https://github.com/ekmett/lens/pull/945. Other than that I think it's good to go.

adamgundry:
Sorry, I haven't been keeping up because of "holidays", but I'm supportive of releasing `optics-0.4` around the GHC-9.0 release if not before. I'll try to find an opportunity to properly review the recent changes in the next week or so. But I don't have anything that is urgent to land before a release.

phadej:
GHC-9.0 is released.

arybczak:
I suggest merging #399 and getting a release out.

adamgundry:
Sounds good to me!

adamgundry:
We don't have CI for GHC 9.0.1 yet (https://github.com/haskell-CI/haskell-ci/issues/474), but I've verified locally that the main packages build and pass tests. 

Still lacking:
 - [ ] `optics-sop` is blocked by https://github.com/well-typed/generics-sop/issues/130 (but isn't released anyway)
 - [ ] `template-haskell-optics` needs to be amended to support `template-haskell-2.17.0.0`
 - [ ] Benchmarks don't build because of dependency bounds, but they can be persuaded to with enough `allow-newer`.

As far as I can see we need to:
 - [ ] Upload 0.4 releases of `optics-core`, `optics-th`, `optics-extra` and `optics` to Hackage.
 - [ ] Relax the bounds of the released `optics-vl` to permit `optics-core-0.4`.
 - [ ] Add release tag.

I'm going to open a PR with some minor changelog fixes. Provisionally I'm planning to make a release early next week. Anything else we need to do first?

adamgundry:
* https://hackage.haskell.org/package/optics-core-0.4/candidate
* https://hackage.haskell.org/package/optics-extra-0.4/candidate
* https://hackage.haskell.org/package/optics-th-0.4/candidate
* https://hackage.haskell.org/package/optics-0.4/candidate

phadej:
[...]

PR #403, opened 2021-02-18
Add changelogs for 0.4 release

adamgundry:
See #375. This adds a few missing changelog entries. Anything else we need to change before the release?

### DIFF
--- a/optics/CHANGELOG.md
+++ b/optics/CHANGELOG.md
@@ -1,3 +1,50 @@
+# optics-0.4 (2021-02-22)
+* See [migration-guide-0.4.md](https://github.com/well-typed/optics/blob/master/migration-guide-0.4.md) for more details
+* Add support for GHC-9.0
+* Drop support for GHC-8.0
+* The `FunctorWithIndex`, `FoldableWithIndex` and `TraversableWithIndex` type
+  classes have been migrated to a new package,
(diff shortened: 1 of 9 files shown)


phadej (commented):
I admit I don't know if there is anything missing.

I don't spot any typos (I never do).

EDIT: I was supposed to click accept.

optics/CHANGELOG.md
```diff
@@ -1,3 +1,31 @@
...
+* Print missing language extensions during TH generation of labels if there are
+  any
+* Add support for getters of rank1 polymorphic fields to optics generated with
+  the `makeFieldLabels` family of functions
+* Extend support of type-changing optics generated with the `makeFieldLabels`
+  family to type parameters that are phantom and applied to non-injective type
+  families
+* Fix TH generation of optics for poly-kinded data families (#378)
+* Fix `declareFieldLabels` when a field type refers to a type defined in the
+  same quote (#380)
```
phadej, line 27: Note, these aren't magically turned into links. Should they be?

arybczak (approved):

optics/CHANGELOG.md
adamgundry: I was being lazy, but now I've done it the brute force way.

adamgundry:
I had missed a few things, so these should be better. I'll release soon.

We should try to maintain changelogs/since pragmas in each PR, rather than rediscovering everything after the fact.

PR #409, opened 2021-03-04
Add 'matching' example

tomjaguarpaw:
and reformat surrounding examples to match

I was really surprised not to find `matching` called out more prominently in the `Prism` Haddock. Could we please add it? Now looks like:

screenshot


adamgundry:
Good idea, thanks for the patch!

PR #410, opened 2021-03-06
Add is

tomjaguarpaw:
(the opposite of isn't)

I found it rather strange that `isn't` exists but `is` doesn't. [`lens` has `is`](https://hackage.haskell.org/package/lens-5.0.1/docs/Control-Lens-Extras.html#v:is). Could we add it to `optics`?

### DIFF
--- a/optics-core/src/Optics/AffineFold.hs
+++ b/optics-core/src/Optics/AffineFold.hs
@@ -150,7 +150,7 @@ infixl 3 `afailing` -- Same as (<|>)
 -- The negation of this operator is 'Optics.Core.Extras.is' from
 -- "Optics.Core.Extras".
 isn't :: Is k An_AffineFold => Optic' k is s a -> s -> Bool
-isn't k s = not (isJust (preview k s))
+isn't k s = isNothing (preview k s)
 {-# INLINE isn't #-}
 
 -- $setup
--- a/optics-core/src/Optics/AffineFold.hs
+++ b/optics-core/src/Optics/AffineFold.hs
@@ -147,6 +147,8 @@ infixl 3 `afailing` -- Same as (<|>)
 -- >>> isn't _Just Nothing
 -- True
 --
+-- The negation of this operator is 'Optics.Core.Extras.is' from
+-- "Optics.Core.Extras".
 isn't :: Is k An_AffineFold => Optic' k is s a -> s -> Bool
 isn't k s = not (isJust (preview k s))
 {-# INLINE isn't #-}
--- a/optics-core/optics-core.cabal
+++ b/optics-core/optics-core.cabal
@@ -88,6 +88,7 @@ library
                    Optics.Operators.Unsafe
                    Optics.Re
                    Optics.ReadOnly
+                   Optics.Core.Extras
 
                    -- optics for data types
                    Data.Either.Optics
--- a/optics-core/src/Optics/Core/Extras.hs
+++ b/optics-core/src/Optics/Core/Extras.hs
@@ -1,0 +1,22 @@
+module Optics.Core.Extras
+  (
+    is
+  )
+  where
+
+import Data.Maybe
+
+import Optics.Optic
+import Optics.AffineFold
+
+-- | Check to see if this 'AffineFold' matches.
+--
+-- >>> is _Just Nothing
+-- False
+--
+is :: Is k An_AffineFold => Optic' k is s a -> s -> Bool
+is k s = isJust (preview k s)
+{-# INLINE is #-}
+
+-- $setup
+-- >>> import Optics.Core


phadej:
`lens` has `is` in the `Control.Lens.Extras` module, which isn't re-exported through `Control.Lens`.

Also `optics` has some `is` in type-signatures. That would be confusing. ;)

I'd rather do nothing, i.e. don't add this combinator.

tomjaguarpaw:
> Also optics has some is in type-signatures. That would be confusing. ;)

I provided the link to lens for informational purposes, not for justification!  The justification was that it was sufficiently confusing to me that `is` does not exist. I imagine that less experienced users would be even *more* confused.

I don't buy the argument that it will be confused with `Is`. It is well-enough established that capitalized identifiers are distinct from uncapitalized ones. However, if you really do hold strongly to that claim can you please memorialise your rationale in the Haddocks as per https://github.com/well-typed/optics/pull/411.

phadej:
I didn't mean that we shouldn't have `is`. I was against exporting it from `Optics.AffinedFold` and thus transitively `Optics.Core` and `Optics`.

And the confusion I was referring to was not with `Is` type family, but with `is` as *plural `i`*. It is a name-clash risky name. Therefore I'd prefer not to have it exported from `Optics` module.

phadej:
@adamgundry, @arybczak what you think about this?

tomjaguarpaw:
OK, thanks for clarifying. Preventing non-export from top level make perfect sense to me. What would be the right way to acheve that?

arybczak:
I'd like `is` to be added to some sort of `Extras` module similar to how `lens` does it.

I recently wanted to use it (with a prism), but had to settle for `has` :(

adamgundry:
I agree that we shouldn't export `is` from `Optics` or `Optics.Core`. I'm not opposed to exporting it from another module if we can think of a name. We have `Optics.Extra` already so we need something different.

Perhaps we should have a module for "things that aren't exported because of name collisions" and then it could also contain `(.) = (%)`...

Although given that `has`/`hasn't` are more general, do we actually gain much from having `is`/`isn't` as well? We could just document the availability of the former?

arybczak:
> We have Optics.Extra already so we need something different.

`Optics.Core.Extras` sounds good to me.

tomjaguarpaw:
OK, how about this? I chose `Optics.Core.Extra` (without an `s`) instead of `Optics.Core.Extras` to match `Optics.Extra`. Let me know if you prefer me to switch that.

Regarding `is`/`isn't`, `has`/`hasn't`, I definitely prefer to have the former because they read better and they guarantee to me that the argument `Is` an `AffineFold` not just a `Fold`, which I think increases legibility. On the other hand if `isn't` is provided then I think it's really important to also provide `is`, to minimise bamboozlement.

tomjaguarpaw:
(I'm not sure why the doctest is failing. I can't reproduce locally due to https://github.com/well-typed/optics/issues/413).

arybczak:
> (I'm not sure why the doctest is failing. I can't reproduce locally due to #413).

Because the file is missing

```
-- $setup
-- >>> import Optics.Core
```

at the end.

optics-core/src/Optics/AffineFold.hs
```diff
@@ -147,6 +147,8 @@ infixl 3 `afailing` -- Same as (<|>)
 -- >>> isn't _Just Nothing
 -- True
 --
+-- The negation of this operator is 'Optics.Core.Extra.is' from
+-- "Optics.Core.Extra".
 isn't :: Is k An_AffineFold => Optic' k is s a -> s -> Bool
 isn't k s = not (isJust (preview k s))
```
arybczak, line 153: This should be `isNothing`. Can you please change that?

optics-core/src/Optics/Core/Extra.hs
```diff
@@ -0,0 +1,17 @@
+module Optics.Core.Extra
+  (
+    is
+  )
+  where
+
+import Optics.Internal.Optic
```
arybczak, line 7: Only `Optics.Optic` is needed, no Internal stuff is used here.

optics-core/src/Optics/Core/Extra.hs
```diff
@@ -0,0 +1,17 @@
...
+import Optics.AffineFold
+
+-- | Check to see if this 'AffineFold' matches.
+--
+-- >>> is _Just Nothing
+-- False
+--
+is :: Is k An_AffineFold => Optic' k is s a -> s -> Bool
+is k = not . isn't k
```
arybczak, line 16: Can you inline the definition of `isn't` here so that the compiler has less things to do?

optics-core/src/Optics/Core/Extra.hs
tomjaguarpaw: Done

optics-core/src/Optics/Core/Extra.hs
tomjaguarpaw: Done

optics-core/src/Optics/AffineFold.hs
tomjaguarpaw: OK

arybczak:
Thanks, looks good to merge to me :+1: Unless @adamgundry @phadej would like the module to have a different name.

adamgundry:
Thanks for taking care of this @tomjaguarpaw!

I'm somewhat torn on the module name question, because `Optics.Extra` exists already as an index of things in the `optics-extra` package, which is rather unrelated to `Optics.Core.Extra`. The latter is most closely related to `Control.Lens.Extras`, of course. I don't have a compelling answer, but I think I'd prefer either `Optics.Core.Extras` (to match `lens`) or perhaps a completely different synonym such as `Optics.Core.Surplus`.

tomjaguarpaw:
OK, I've changed the module name to `Optics.Core.Extras`.

arybczak:
Thanks :) I'll merge after the weekend unless anyone objects.

PR #411, opened 2021-03-06
Explain why "is" is not provided

tomjaguarpaw:


tomjaguarpaw:
If the reason for not providing "is" is that there would be a confusing clash of terminology with "Is" then could you please call that out explicitly? Otherwise I fear that users will think there is a deep technical reason it is not provided. It certainly puzzled me for a good few minutes. The world of optics is confusing enough already. Please be explicit!

tomjaguarpaw:
(Personally I prefer https://github.com/well-typed/optics/pull/410 but if you really thin that `is` shouldn't be provided then explaining the justification will go a long way to unpuzzling users.)

tomjaguarpaw:
Closing since this was due to a misunderstanding.

PR #417, opened 2021-04-02
Release indexed-profunctors-0.1.1

arybczak:


arybczak:
I'll upload a release to hackage during the weekend if there are no objections.

PR #422, opened 2021-05-29
I'm trying out haskell-ci ghcup and GHC-9.2 support

phadej:


phadej:
Looks like it works. Good.

Issue #308, opened 2020-04-09
Backreasoning (with holes).


phadej:
The example (by @kosmikus)

```haskell
{-# LANGUAGE TemplateHaskell #-}

import Optics

data Person =
  MkPerson
    { _name :: String
    , _numbers :: (Int, Int)
    }

makeLenses ''Person

p :: Person
p = MkPerson "Frodo" (66, 88)

example :: Person
example = set (numbers % _) 33 p
```

causes far from optimal type errors:

```
src/LvO.hs:17:11: error:
    • Overlapping instances for Is (Join A_Lens l0) A_Setter
        arising from a use of ‘set’
      Matching instances:
        instance [overlappable] (TypeError ...) => Is k l
          -- Defined in ‘optics-core-0.2:Optics.Internal.Optic.Subtyping’
        instance Is k k
          -- Defined in ‘optics-core-0.2:Optics.Internal.Optic.Subtyping’
        instance Is A_Lens A_Setter
          -- Defined in ‘optics-core-0.2:Optics.Internal.Optic.Subtyping’
        ...plus four others
        (use -fprint-potential-instances to see them all)
      (The choice depends on the instantiation of ‘l0’
       To pick the first instance above, use IncoherentInstances
       when compiling the other instance declarations)
    • In the expression: set (numbers % _) 33 p
      In an equation for ‘example’: example = set (numbers % _) 33 p
   |
17 | example = set (numbers % _) 33 p
   |           ^^^^^^^^^^^^^^^^^^^^^^

src/LvO.hs:17:16: error:
    • Overlapping instances for Is l0 (Join A_Lens l0)
        arising from a use of ‘%’
      Matching instances:
        instance [overlappable] (TypeError ...) => Is k l
          -- Defined in ‘optics-core-0.2:Optics.Internal.Optic.Subtyping’
        instance Is k k
          -- Defined in ‘optics-core-0.2:Optics.Internal.Optic.Subtyping’
        instance Is A_Getter A_Fold
          -- Defined in ‘optics-core-0.2:Optics.Internal.Optic.Subtyping’
        ...plus 35 others
        (use -fprint-potential-instances to see them all)
      (The choice depends on the instantiation of ‘l0’
       To pick the first instance above, use IncoherentInstances
       when compiling the other instance declarations)
    • In the first argument of ‘set’, namely ‘(numbers % _)’
      In the expression: set (numbers % _) 33 p
      In an equation for ‘example’: example = set (numbers % _) 33 p
   |
[...]

PR #424, opened 2021-06-08
Improve error message for impossible JoinKinds constraint

### LINKED ISSUE #423: Misleading error message from JoinKinds
With 0.4 and current HEAD:
```
ghci> :t castOptic @A_Setter @A_Setter (traversed % _1)

<interactive>:1:42: error:
    • A_Traversal cannot be composed with A_Lens
    • In the third argument of ‘castOptic’, namely ‘(traversed % _1)’
      In the expression: castOptic @A_Setter @A_Setter (traversed % _1)
```

Oops! I think what is going on here is that we get a constraint `JoinKinds A_Traversal A_Lens A_Setter` which of course does not match the instance `JoinKinds A_Traversal A_Lens A_Traversal`, so the catch-all error instance fires. Ideally we would have two different errors, one for "this composition doesn't make sense" and another for "this composition makes sense, but produces an optic of a different kind to that expected by the context".

adamgundry:
Fixes #423.

WARNING: typeclass Prolog ahead.

### DIFF
--- a/optics-core/src/Optics/Internal/Optic/Subtyping.hs
+++ b/optics-core/src/Optics/Internal/Optic/Subtyping.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE TypeInType #-}
 {-# LANGUAGE UndecidableInstances #-}
@@ -467,7 +468,7 @@ instance {-# OVERLAPPABLE #-}
 type family CannotJoinKinds k l m actual is_concrete :: Constraint where
   CannotJoinKinds k l m actual is_concrete =
     ( -- 1. Unify is_concrete with True or False
-      IsOpticKind m is_concrete
+      IsConcrete m is_concrete
       -- 2. If is_concrete, unify actual with the join of k and m (otherwise
       -- do nothing; the conditional is necessary to avoid an infinite
       -- constraint solver loop).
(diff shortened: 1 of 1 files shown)


arybczak:
Thanks. Do we want to drop support for GHC 8.2 and 8.4 (as this introduces `UndecidableSuperClasses`)? I personally don't mind.

Also, doctests fail, presumably due to a missing import.

optics-core/src/Optics/Internal/Optic/Subtyping.hs
```diff
@@ -432,8 +444,80 @@ instance JoinKinds A_Setter           A_Traversal        A_Setter           wher
...
+      ':$$: 'Text "but the context expects " ':<>: 'ShowType m
+  CannotJoinKindsMessage k l m actual 'False
+    = 'ShowType k ':<>: 'Text " cannot be composed with " ':<>: 'ShowType l
+
+-- | When solving an @IsOpticKind m v@ constraint, if @m@ is known to be an
+-- optic kind then @is_concrete@ will be unified with True; otherwise
+-- @is_concrete@ will be unified with @False@.
+--
+class IsOpticKind (m :: OpticKind) (is_concrete :: Bool) | m -> is_concrete
+instance is_concrete ~ 'True => IsOpticKind An_AffineFold      is_concrete
```
phadej, line 510: What will happen if a kind is forgotten, we get a "wrong" custom type error?

adamgundry: Yes, I was a bit dissatisfied with the need to list all the optic kinds explicitly, but I couldn't immediately see how to avoid it. Having thought a bit more I've got a version that doesn't need this.

adamgundry:
I was playing around with this branch and noticed this:
```
> view @A_Getter (folded % _) [Just "abc", Nothing, Just "def"]

<interactive>:3:17: error:
    • A_Fold composed with l0 produces p0,
      but the context expects A_Getter
```
We should be able to do better here, by checking if more parameters are concrete.

On the other hand, we're a bit stuck if the context doesn't force a particular optic kind (cf. #308):
```
> view (folded % _) [Just "abc", Nothing, Just "def"]

<interactive>:4:1: error:
    • Overlapping instances for Is k0 A_Getter
        arising from a use of ‘view’
...
    • Overlapping instances for JoinKinds A_Fold l0 k0
        arising from a use of ‘%’
```

arybczak:
What's the status of this PR?

adamgundry:
I've been away on holiday so apologies for the radio silence. I'm still planning to polish this a bit, but it may be a little while before I can get back to it.

arybczak:
@adamgundry I posted an alternative solution in #439, please have a look if you have a moment. It's not as fancy, but should be more predictable :)

arybczak:
Closing since #439 was merged a long time ago and it more or less supersedes it.

Issue #426, opened 2021-06-11
Improve Optics.Fold docs


adamgundry:
At the moment `Optics.Fold` uses `foldVL` as the canonical introduction form, and has lots of canonical elimination forms. I think it would be clearer to use `folding @[]` instead for the introduction form, and `toListOf` for the elimination form. This makes explicit the connection between `Fold s a` and `s -> [a]`.

arybczak:
IMO `foldVL_` is much easier to grok than `folding` :thinking: Its resemblence to `traversalVL` is analogous to how `traverse_` and `traverse` differ.

adamgundry:
You're right that we should emphasise the relationship between `traversalVL` and `foldVL`. That's helpful if one understands traversals already. But I think that as a route to understanding folds without assuming an understanding of traversals, `s -> [a]` is simpler.

I'm currently working on a workshop about `optics` for ZuriHac, which is why I'm thinking about this. Once that is under control I'll try to make a PR with the doc changes I envision.

phadej:
`foldring` with a stricter signature, i.e.

```haskell
(forall b. (a -> b -> b) -> b -> s -> b) -> Fold s a
```

is IMHO simplest introduction form. Often quite optimal as well.

As I hint in #427, our type for `foldVL` is not entirely correct - and we don't discuss that in the docs:

```haskell
*Optics.Core> :t foldVL traverse
foldVL traverse :: Traversable t => Fold (t a) a
```

works, but i'm not sure it should or shouldn't. (`folding foldMap` of `foldring foldr` would be better, wouldn't it?)

arybczak:
> folding foldMap of foldring foldr would be better, wouldn't it?

These provide different performance characteristics. Last time I checked `foldVL traverse_` was the fastest for most standard containers.

Issue #179, opened 2019-06-01
mkFold vs foldVL


arybczak:
Split from #165.

`mkFold` used to be called `foldVL`, but I changed it as it takes something more general than VL fold since `f` doesn't need to be `Contravariant`.

But it occurs to me now that we can't have proper `foldVL` because we use `wander` for folds internally and it doesn't know about `Contravariant`.

Shall we rename it back? I'd be inclined to do so. @adamgundry what do you think?

adamgundry:
As a point of comparison, we have `atraversalVL` and `toAtraversalVL` for `AffineTraversal`, even though our definition of `AffineTraversalVL` is slightly wrong as it takes the "`Pointed` dictionary" as an explicit argument. So there's precedent for deviating slightly here. As long as we document the infelicity I'd be happy with `foldVL`.

PR #427, opened 2021-06-11
Emphasize similarities between traversalVL and foldVL

arybczak:
Somewhat related to #426.

### DIFF
--- a/optics-core/src/Optics/Traversal.hs
+++ b/optics-core/src/Optics/Traversal.hs
@@ -124,13 +124,18 @@ type TraversalVL s t a b = forall f. Applicative f => (a -> f b) -> s -> f t
 -- | Type synonym for a type-preserving van Laarhoven traversal.
 type TraversalVL' s a = TraversalVL s s a a
 
--- | Build a traversal from the van Laarhoven representation.
+-- | Construct a 'Traversal' from a 'traverse' like function.
+--
+-- /Note:/ for lifting a 'Data.Foldable.traverse_' like function see
+-- 'Optics.Fold.foldVL'.
 --
 -- @
 -- 'traversalVL' '.' 'traverseOf' ≡ 'id'
 -- 'traverseOf' '.' 'traversalVL' ≡ 'id'
 -- @
-traversalVL :: TraversalVL s t a b -> Traversal s t a b
+traversalVL
+  :: (forall f. Applicative f => (a -> f b) -> s -> f t)
+  -> Traversal s t a b
 traversalVL t = Optic (wander t)
 {-# INLINE traversalVL #-}
 
--- a/optics-core/src/Optics/Fold.hs
+++ b/optics-core/src/Optics/Fold.hs
@@ -117,14 +117,17 @@ import Optics.Internal.Utils
 -- | Type synonym for a fold.
 type Fold s a = Optic' A_Fold NoIx s a
 
--- | Obtain a 'Fold' by lifting 'traverse_' like function.
+-- | Construct a 'Fold' from a 'traverse_' like function.
+--
+-- /Note:/ for lifting a 'Data.Traversable.traverse' like function see
+-- 'Optics.Traversal.traversalVL'.
 --
 -- @
 -- 'foldVL' '.' 'traverseOf_' ≡ 'id'
 -- 'traverseOf_' '.' 'foldVL' ≡ 'id'
 -- @
 foldVL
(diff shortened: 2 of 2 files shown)


optics-core/src/Optics/Traversal.hs
```diff
@@ -124,13 +124,18 @@ type TraversalVL s t a b = forall f. Applicative f => (a -> f b) -> s -> f t
...
+-- /Note:/ for lifting a 'Data.Foldable.traverse_' like function see
+-- 'Optics.Fold.foldVL'.
 --
 -- @
 -- 'traversalVL' '.' 'traverseOf' ≡ 'id'
 -- 'traverseOf' '.' 'traversalVL' ≡ 'id'
 -- @
-traversalVL :: TraversalVL s t a b -> Traversal s t a b
+traversalVL
+  :: (forall f. Applicative f => (a -> f b) -> s -> f t)
```
phadej, line 137: I don't like this. We have unfolded `FoldVL` definition for some reason (and it's not correct). Yet we very much could, as `Contravariant` is in base since 4.12 (we'd need to drop ghc-8.2 support, that's imo fine). We also use synonyms in `optics-vl` and in `Optics.Lens`.

arybczak: What do you mean by "it's not correct"?

arybczak: Ok, I think you mean that it's not how `Fold` from `lens` looks like.

This is fine. `Contravariant` in `Fold` from `lens` is a quirk emerging from their implementation. We don't need `Contravariant` because we have `Bicontravariant` internally to do the same thing.

You mentioned in #426 that `foldVL traverse` works, but that's just because `foldVL` type is too permissive. I actually changed it here so that `foldVL traverse` no longer works, but `foldVL traverse_` works just fine.

arybczak:
~~I'd like to merge this. Any objections?~~

Closing, it's not very relevant after #430.

optics-core/src/Optics/Traversal.hs
phadej: Not fine. I cannot convert `lens` `Fold` into `optics` `Fold` using current, or new `foldVL`.

```haskell
import qualified Control.Lens as L
import qualified Optics as O

ex1 :: Foldable f => L.Fold (f a) a
ex1 = L.folding id

ex2 :: Foldable f => O.Fold (f a) a
ex2 = O.foldVL ex1

{-

fold.hs:8:16: error:
    • Could not deduce (L.Contravariant f1) arising from a use of ‘ex1’
      from the context: Foldable f
        bound by the type signature for:
                   ex2 :: forall (f :: * -> *) a. Foldable f => O.Fold (f a) a
        at fold.hs:7:1-35
      or from: Applicative f1
        bound by a type expected by the context:
                   forall (f1 :: * -> *).
                   Applicative f1 =>
                   (a -> f1 a) -> f a -> f1 (f a)
        at fold.hs:8:7-18
      Possible fix:
        add (L.Contravariant f1) to the context of
          a type expected by the context:
            forall (f1 :: * -> *).
            Applicative f1 =>
            (a -> f1 a) -> f a -> f1 (f a)
    • In the first argument of ‘O.foldVL’, namely ‘ex1’
      In the expression: O.foldVL ex1
      In an equation for ‘ex2’: ex2 = O.foldVL ex1
  |
8 | ex2 = O.foldVL ex1

-}
```

arybczak: I mentioned at least once (https://github.com/well-typed/optics/issues/179 and https://github.com/well-typed/optics/pull/183) before the initial release the discrepancy you're pointing and we agreed it was fine.

In any case `foldVL` can't "easily" take `Contravariant` because `wander` doesn't.

It seems we just need to communicate the fact that `foldVL` is not in compatible with `lens`. I had a patch lying around, so if you really want this, I rebased and pushed it: #430.

phadej: There's not much discussion in #183, and in light of new knowledge I think it wasn't that good of the move. `mkFold` is not great name either, but `foldVL` is simply wrong and with #430 in it would be misleading and confusing.

arybczak: I'd like to resolve this situation, otherwise these two PRs (this and #430) will keep hanging.

Let's just rename `foldVL` in `Optics.Fold` back to `mkFold`. Yes, it's not a great name, but it's better to rename it than be in the current situation.

Unless you have propositions for a better name :slightly_smiling_face:

PR #430, opened 2021-06-25
Add support for converting to/from VL getters and folds

### LINKED ISSUE #533: Add getterVL
[A user was confused](https://discourse.haskell.org/t/another-lens-optics-type-difference/13335?u=tomjaguarpaw) why `lensVL` didn't work on a `Control.Lens.Getter`. The answer is, of course, because `lensVL` expects a `Lens`, not a `Getter`. So the user should use `getterVL` instead. But it doesn't exist! Is there any reason why it couldn't or shouldn't?

arybczak:
Fixes #533 and #535.

### DIFF
--- a/indexed-profunctors/CHANGELOG.md
+++ b/indexed-profunctors/CHANGELOG.md
@@ -1,3 +1,6 @@
+# indexed-profunctors-0.1.2.0 (2026-??-??)
+* Add a `Cochoice` instance for `Star`.
+
 # indexed-profunctors-0.1.1.1 (2023-06-22)
 * Add INLINE pragmas to small functions that really should inline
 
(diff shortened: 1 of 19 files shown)


optics-vl/src/Optics/VL.hs
```diff
@@ -168,3 +216,33 @@ withPrismVL
...
+{-# INLINE toGetterVL #-}
+
+----------------------------------------
+
+type FoldVL s a =
+  forall f. (Contravariant f, Applicative f) => (a -> f a) -> s -> f s
+
+-- | Build a 'Fold' from the van Laarhoven representation.
+foldVL :: FoldVL s a -> Fold s a
+foldVL o = O.foldVL $ \f -> runTraversed . getConst #. o (Const #. Traversed #. f)
```
phadej, line 242: This looks great. But how to rename `foldVL` we already have. I don't have any ideas.

arybczak: foldVL used to be named `mkFold`. How about that?

PR #434, opened 2021-10-04
Add `(%?)` composition operator

matobet:
```haskell
o1 %? o2 = o1 % _Just % o2
```
Useful helper when "digging" through lot of Maybe fields, mirroring the usage of `?.` operator from some object-oriented languages.

### DIFF
--- a/optics-core/src/Data/Maybe/Optics.hs
+++ b/optics-core/src/Data/Maybe/Optics.hs
@@ -6,9 +6,11 @@
 module Data.Maybe.Optics
   ( _Nothing
   , _Just
+  , (%?)
   )
   where
 
+import Optics.Internal.Optic
 import Optics.Prism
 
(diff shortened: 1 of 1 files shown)


arybczak:
@matobet please fix typos in the haddock.

phadej:
> I've found that this gets used quite a lot, I'm inclined to include it in the library. ... thoughts?

Fine with me. I'm not a fan of new operators in general.

I looked at a lens-heavy codebase and while `somelens . _Just . otherlens` is common, also `lens . _Just` and `_Just . lens` are also common, even cases like: `p ^? _Just . oneRecordField . _Just`, and there `_Just` would still persist.

I don't know whether I would remember to use `%?`, maybe if HLint warned.

matobet:
@arybczak done. Hopefully caught them all

PR #439, opened 2021-12-06
Stop picking the catch-all JoinKinds instance if arguments match

### LINKED ISSUE #423: Misleading error message from JoinKinds

arybczak:
Fixes #423 (to an extent).

Simpler than #424, but should make the situation strictly better.

### DIFF
--- a/codegen/Subtypes.hs
+++ b/codegen/Subtypes.hs
@@ -117,7 +117,7 @@ genJoin = either (fail . show) id $ runG opticsKind $ \g -> do
             case mkl of
                 -- alignment is important, thus padding
                 []          -> putStrLn $ unwords
-                  [ "--    no JoinKinds"
+                  [ "--                              no JoinKinds"
                   , leftpad (show k)
                   , leftpad (show l)
                   ]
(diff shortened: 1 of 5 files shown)


arybczak:
Uh, the fragility of these inspection tests is rather sad.

I also removed the `let res :: JoinKinds A_Lens A_Prism k => Proxy k; res = Proxy` doctest from the `Optics` module because it started triggering a warning with this PR. It's fixable, but it no longer works as expected with GHC 9.2 anyway.

codegen/Subtypes.hs
```diff
@@ -126,10 +126,12 @@ genJoin = either (fail . show) id $ runG opticsKind $ \g -> do
...
   where
     joinInstance k l m = unwords
-      [ "instance JoinKinds"
+      [ "instance k ~"
+      , leftpad $ show m
+      , "=> JoinKinds"
       , leftpad $ show k
       , leftpad $ show l
-      , leftpad $ show m
+      , "k"
```
phadej, line 134: Clever :)

phadej (approved):

PR #440, opened 2021-12-09
Compile benchmarks and run doctest with more GHC versions

arybczak:


arybczak:
Uh, what's up with 9.2.1 :thinking:

arybczak:
Failure with 9.0.1 is a GHC bug: https://gitlab.haskell.org/ghc/ghc/-/issues/19460

PR #443, opened 2022-02-08
add parallel optics

oberblastmeister:
adds `parOver`, `parOf`, and `evalOf`

### DIFF
--- a/optics-extra/src/Control/Parallel/Strategies/Optics.hs
+++ b/optics-extra/src/Control/Parallel/Strategies/Optics.hs
@@ -1,0 +1,42 @@
+module Control.Parallel.Strategies.Optics
+  ( parOver
+  , evalOf
+  , parOf
+  ) where
+
+import qualified Control.Parallel.Strategies as Parallel
+import Optics.Core
+
+-- | Apply a traversal as a modifier in parallel.
+parOver :: Is k A_Traversal => Parallel.Strategy b -> Optic k is s t a b -> (a -> b) -> s -> t
+parOver strat o f = Parallel.runEval . traverseOf o (Parallel.rparWith strat . f)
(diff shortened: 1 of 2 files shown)


arybczak:
I think this is fine. @phadej any thoughts?

optics-extra/src/Control/Parallel/Strategies/Optics.hs
```diff
@@ -0,0 +1,42 @@
...
+-- | Evaluate the targets of a 'Lens' or 'Traversal' into a data structure
+-- according to the given 'Strategy'.
+--
+-- @
+-- 'evalTraversable' = 'evalOf' 'traverse' = 'traverse'
+-- 'evalOf' = 'id'
+-- @
+--
+-- @
+-- 'evalOf' :: 'Lens'' s a -> 'Strategy' a -> 'Strategy' s
```
arybczak, line 24: We don't need to add these as `lens` does since our type signatures are clearer.

optics-extra/src/Control/Parallel/Strategies/Optics.hs
```diff
@@ -0,0 +1,42 @@
...
+evalOf = traverseOf
+{-# INLINE evalOf #-}
+
+-- | Evaluate the targets of a 'Lens' or 'Traversal' according into a
+-- data structure according to a given 'Strategy' in parallel.
+--
+-- @'parTraversable' = 'parOf' 'traverse'@
+--
+-- @
+-- 'parOf' :: 'Lens'' s a -> 'Strategy' a -> 'Strategy' s
```
arybczak, line 37: Same as above.

arybczak:
Sounds reasonable. @oberblastmeister please make a separate package for these if you're interested.

PR #446, opened 2022-03-08
Update documentation of Optics.Label

arybczak:
@adamgundry can you please confirm that the update is accurate?


adamgundry (approved):
Thanks for updating this, it looks right to me.

PR #451, opened 2022-03-22
Better inference for AppendIndices

treeowl:
* Allow the second argument of `AppendIndices` to be determined by the first and third arguments.
* Refine the documentation for `AppendIndices`. Due to the way incoherent instances work, the "short-circuit" path is actually treated as a fall-back; it will *only* be taken if nothing is known about the shape of the first list.

### DIFF
--- a/optics-core/src/Optics/Internal/Optic/TypeLevel.hs
+++ b/optics-core/src/Optics/Internal/Optic/TypeLevel.hs
@@ -111,12 +111,12 @@ data IxEq i is js where
 class AppendIndices xs ys ks | xs ys -> ks where
   appendIndices :: IxEq i (Curry xs (Curry ys i)) (Curry ks i)
 
--- | If the second list is empty, we can shortcircuit and pick the first list
--- immediately.
+-- | If the second list is empty, we can pick the first list
+-- even if nothing is known about it.
 instance {-# INCOHERENT #-} AppendIndices xs '[] xs where
   appendIndices = IxEq
 
-instance AppendIndices '[] ys ys where
+instance ys ~ zs => AppendIndices '[] ys zs where
   appendIndices = IxEq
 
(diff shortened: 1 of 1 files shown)


optics-core/src/Optics/Internal/Optic/TypeLevel.hs
```diff
@@ -111,12 +111,12 @@ data IxEq i is js where
 class AppendIndices xs ys ks | xs ys -> ks where
   appendIndices :: IxEq i (Curry xs (Curry ys i)) (Curry ks i)
 
--- | If the second list is empty, we can shortcircuit and pick the first list
--- immediately.
+-- | If the second list is empty, we can pick the first list
+-- even if nothing is known about it.
 instance {-# INCOHERENT #-} AppendIndices xs '[] xs where
```
arybczak, line 116: How about making this one similar to the one below?

```suggestion
instance {-# INCOHERENT #-} xs ~ zs => AppendIndices xs '[] zs where
```
?

arybczak (commented):
Do you have an example to demonstrate how this change helps?

optics-core/src/Optics/Internal/Optic/TypeLevel.hs
treeowl: That doesn't look like a bad idea either.

treeowl:
> Do you have an example to demonstrate how this change helps?

I don't know enough about the package to say, really, but the main idea is that once the length of the first list is known, we can learn about the second list by looking at the third. So in particular, if we see that the third list is curried a certain number of times, and applied to arguments of certain types, we can fill in parts of the second list, possibly enabling other things to be inferred.

phadej:
There is indeed `ks xs -> ys` and `ks ys -> xs` injectivity, but we cannot convince GHC about either, IIRC (we could about first one if there weren't third incoherent instance?). It's good if this change brings us closer to that though.

PR #460, opened 2022-05-20
Mention the result in the error message for JoinKinds

### LINKED ISSUE #453: Error message for `JoinKinds` could mention m
Suggested by a user and passed on by @kosmikus. The error message from the overlappable `JoinKinds k l m` instance currently says `k` cannot be composed with `l`, but doesn't mention `m`. At least when `m` is a concrete type it seems like it would be nice to mention it. Related to #423.

arybczak:
Fixes #453.

I avoided any fancy tricks to make it appear only if `m` is concrete for simplicity.


arybczak:
The doctest error is utterly bizarre.

```
src/Optics.hs:540: failure in expression `:t mapped % folded'
expected: ...
          ...A_Setter cannot be composed with A_Fold
          ...
 but got: mapped % folded
            :: ((TypeError ...), Functor f1, Foldable f2) =>
               Optic m '[] (f1 (f2 b)) (f1 (f2 b)) b b
           ^
```

Looks like another TypeError GHC quirk :disappointed:

adamgundry:
That is bizarre. It looks like GHC is successfully inferring a type for `mapped % folded`, albeit one that includes a `TypeError`, and is then printing the inferred type rather than firing the `TypeError`.  At a guess, perhaps the presence of `m` in the `TypeError` makes GHC quantify over the constraint?

adamgundry:
The issue seems to be that with the `:t` command, if the type of the expression is polymorphic, any `TypeError` constraints are quantified over rather than being reported as errors. For example:
```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE TypeFamilies #-}

import GHC.TypeLits

type family F a where
  F a = TypeError (ShowType a)

-- :t plugh reports a type error
plugh :: F Int => Int -> ()
plugh _ = ()

-- :t blah doesn't report a type error
blah :: F a => a -> ()
blah _ = ()

-- :t bloo doesn't report a type error
bloo :: Show (F Int) => a -> ()
bloo _ = ()
```

```
GHCi, version 9.2.2: https://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( TypeErrorQuantification.hs, interpreted )
Ok, one module loaded.
ghci> :t plugh

<interactive>:1:1: error:
    • Int
    • In the expression: plugh
ghci> plugh

<interactive>:2:1: error:
    • Int
    • In the expression: plugh
      In an equation for ‘it’: it = plugh
ghci> :t blah
blah :: (TypeError ...) => a -> ()
ghci> :t \ (x :: a) -> blah @a
\ (x :: a) -> blah @a :: (TypeError ...) => a -> a -> ()
ghci> :t blah @Int

<interactive>:1:1: error:
    • Int
    • In the expression: blah @Int
ghci> blah

<interactive>:6:1: error:
    • a
        it :: forall {a}. (TypeError ...) => a -> ()
ghci> :t bloo
bloo :: Show (TypeError ...) => a -> ()
```

This seems very inconsistent, and is more motiviation for `Unsatisfiable` at least. But it might also be worth a GHC ticket for `TypeError`?

arybczak:
Ok, resurrected.

Funnily enough the TypeError underreporting problem was fixed in GHC >= 9.4, but Unsatisfiable has it even though AFAIR it was supposed to be more eager than TypeError, not less (I tried resurrecting #529 and stumbled upon it).

I'll make a GHC bug report about Unsatisfiable.

PR #463, opened 2022-05-31
Test with GHC 9.2.3

arybczak:


phadej:
https://gitlab.haskell.org/haskell/ghcup-hs/-/issues/370

If you add `distribution: focal` or `distribution: jammy`  then it should work. (default `bionic` is there for older GHCs. I'll bump default ditribution to `focal` in `haskell-ci` when `bionic` is EOLd, which is soon but not just yet).

Issue #237, opened 2019-09-03
Module links not working on hackage


barischrooneyj:
![Screenshot 2019-09-03 at 16 36 14](https://user-images.githubusercontent.com/6631452/64185849-1d037100-ce6e-11e9-8875-c0745cdbb0c6.png)

adamgundry:
Yes, this is a consequence of Hackage not supporting `reexported-modules` very well unfortunately (see haskell/hackage-server#577). You can see the links in `optics-core` and `optics-extra`, but the reexports in `optics` don't get linked. I'm not sure if we can do much to improve matters...

adamgundry:
We have a workaround for this in #465, but the script needs to be run and the results uploaded for each release, and it seems that wasn't done for the most recent release.

A worry is that `@since` pragmas in re-exported modules will be copied verbatim by the script, even though they apply to (e.g.) `optics-core` rather than `optics`.

phadej:
> optics-core rather than optics.

That's a small worry, as `optics(-core)` versions are synchronized. If something is *since `optics-core-0.5.6` it won't be in `optics-0.4`, but only in `optics-0.5.6`. In particular, there will be `optics-0.5.6` to correspond to `optics-core-0.5.6` if the addition is in the re-exported module (and in particular visible in haddocks).

PR #465, opened 2022-07-06
haddock bundle script

phadej:


adamgundry:
Nice! Thanks for looking into this. I haven't looked extremely closely at the code, but the results on https://hackage.haskell.org/package/optics look good. :smile: 

Did you manage to resolve the issue with there being a few missing links?

We should remove the note from the description saying "Hackage does not yet display documentation for reexported-modules".

Fixes #237.

arybczak:
It seems that modules from `optics-extra` are missing links :disappointed:

arybczak:
I wonder if providing dummy files that redirect to modules in appropriate package with `<meta http-equiv="refresh" content="0; url=...>` would work :thinking:

phadej:
> Did you manage to resolve the issue with there being a few missing links?

No, that seems to be some issue with `haddock`, i have no idea what it could be.

haddockbundle.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+
```
arybczak, line 2: Can you add a brief "Usage" section? Even if it's running it without any arguments.

phadej:
I added a commit which seems to fix links from `optics-extra` packages

adamgundry:
Nice work tracking that down! And sorry for letting this linger, let's get it merged. I'm assuming you already uploaded docs to Hackage, since I see them published...

I think we should:
 * Remove the note from the description saying "Hackage does not yet display documentation for reexported-modules".
 * Briefly document the release process somewhere. Is the only new step is running `haddockbundle.sh` and uploading the generated `dist-newstyle/optics-bundled-docs.tar`, and can one do that directly with `cabal upload -d`?

phadej:
>  can one do that directly with cabal upload -d

I don't know, never used that one.

> Briefly document the release process somewhere.

Once someone does it, they could write a trace of what they did.

adamgundry (approved):

PR #466, opened 2022-08-12
Test with GHC 9.4.1

arybczak:
9.4.1 job needs https://github.com/well-typed/generics-sop/pull/156.


adamgundry (approved):
I re-ran the failing test now that @phadej has updated `generics-sop`, so this looks good to go now.

PR #475, opened 2022-12-28
Restrict over', iover', and set' to traversals

### LINKED ISSUE #473: over' and iover' pretend to work with incompatible optics
```
ghci> length $ over' mapped id [undefined, undefined]
2
```

Oops. I'm currently working on adding `over'`, `iover'`, and relatives to `lens`. The choice I've made over there is *not* to add a `Settable Solo` instance, to avoid this very problem. I don't know how these optics work, but the uses of `unwrapIdentity'` in the `Mapping (Star Identity')` and `Mapping (IxStar Identity')` instances look really fishy. What goes wrong if those instances are removed, and `over'` and `iover'` are declared to need traversals rather than setters?

Additional surprising uses:

`failover'` and `ifailover'` both do things like

```haskell
if visited
  then Just (unwrapIdentity' t)
  else Nothing
```

Should those be like this?

```haskell
if | visited
   , Identity' () v <- t
   -> Just v
   | otherwise
   -> Nothing
```

or (using `Solo`)

```haskell
if | visited
   , Solo v <- t
   = Just v
   | otherwise
   = Nothing
```

----

While you're in the ballpark, `Identity'` is quite an awkward type, and there's really no need for such complication.

```haskell
Identity'  ~=  Solo -- Best Solo source is OneTuple, which handles all the compat mess
unwrapIdentity' ~= getSolo
wrapIdentity' ~= (Solo $!)
```

treeowl:
* `over'`, `iover'`, `set'`, and associated operators previously
  accepted setters. However, it's impossible to actually modify strictly
  through a setter; a traversal is needed for that. Restrict the types
  to require `A_Traversal`, and remove the associated (technically
  correct but deceptive) `Mapping` instances.

* Document the strictness behavior of `set'`.

* The internal `Identity'` type was non-standard and overly complicated.
   The `Solo` type, now in `base`, is operationally identical and simpler to
  deal with. Use that instead.

* Remove some surprisingly lazy matches on `Identity'`/`Solo`.

Fixes #473

### DIFF
--- a/optics-core/optics-core.cabal
+++ b/optics-core/optics-core.cabal
@@ -77,6 +77,7 @@ library
                , indexed-profunctors    >= 0.1        && <0.2
                , transformers           >= 0.5        && <0.7
                , indexed-traversable    >= 0.1        && <0.2
+               , OneTuple               >= 0.3        && <0.4
 
   exposed-modules: Optics.Core
 
--- a/optics-core/src/Optics/Setter.hs
+++ b/optics-core/src/Optics/Setter.hs
@@ -63,10 +63,10 @@ module Optics.Setter
   ) where
 
 import Data.Profunctor.Indexed
+import Data.Tuple.Solo (Solo (..), getSolo)
 
 import Optics.Internal.Optic
 import Optics.Internal.Setter
-import Optics.Internal.Utils
 
 -- | Type synonym for a type-modifying setter.
 type Setter s t a b = Optic A_Setter NoIx s t a b
@@ -102,13 +102,27 @@ over o = \f -> runFunArrow $ getOptic (castOptic @A_Setter o) (FunArrow f)
 -- 'over' is used, because the first coordinate of a pair is never forced.
 --
 over'
-  :: Is k A_Setter
+  :: Is k A_Traversal
   => Optic k is s t a b
   -> (a -> b) -> s -> t
+-- See [Note: Solo wrapping]
 over' o = \f ->
-  let star = getOptic (castOptic @A_Setter o) $ Star (wrapSolo' . f)
+  let star = getOptic (castOptic @A_Traversal o) $ Star ((Solo $!) . f)
   in getSolo . runStar star
 {-# INLINE over' #-}
+
+-- Note: Solo wrapping
+--
+-- We use Solo for strict application of (indexed) setters.
+--
+-- Credit for this idea goes to Eric Mertens; see
+-- <https://github.com/glguy/irc-core/commit/2d5fc45b05f1>.
+--
+-- Using Solo rather than Identity allows us, when applying a traversal to a
+-- structure, to evaluate only the parts that we modify. If an optic focuses on
+-- multiple targets, the Applicative instance of Solo (combined with applying
+-- the Solo data constructor strictly) makes sure that we force evaluation of
+-- all of them, but we leave anything else alone.
 
 -- | Apply a setter.
 --
@@ -128,10 +142,11 @@ set o = over o . const
 
 -- | Apply a setter, strictly.
 --
--- TODO DOC: what exactly is the strictness property?
---
+-- The new value will be forced if and only if the optic traverses at
+-- least one target. If forcing the new value is inexpensive, then it
+-- is cheaper to do so manually and use 'set'.
 set'
-  :: Is k A_Setter
+  :: Is k A_Traversal
   => Optic k is s t a b
   -> b -> s -> t
 set' o = over' o . const
--- a/optics-core/src/Optics/Setter.hs
+++ b/optics-core/src/Optics/Setter.hs
@@ -106,8 +106,8 @@ over'
   => Optic k is s t a b
   -> (a -> b) -> s -> t
 over' o = \f ->
-  let star = getOptic (castOptic @A_Setter o) $ Star (wrapIdentity' . f)
-  in unwrapIdentity' . runStar star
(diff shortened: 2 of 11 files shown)


treeowl:
I think it would be rather nice to add the `strictly` combinator I've [proposed for `lens`](https://github.com/ekmett/lens/pull/1017), but it would probably take me a bit to learn enough about how this package works to be able to do that myself.

optics-core/src/Optics/Setter.hs
```diff
@@ -102,14 +102,28 @@ over o = \f -> runFunArrow $ getOptic (castOptic @A_Setter o) (FunArrow f)
...
-  :: Is k A_Setter
+  :: Is k A_Traversal
   => Optic k is s t a b
   -> (a -> b) -> s -> t
+-- See [Note: Solo wrapping]
 over' o = \f ->
-  let star = getOptic (castOptic @A_Setter o) $ Star (wrapIdentity' . f)
-  in unwrapIdentity' . runStar star
+  let star = getOptic (castOptic @A_Traversal o) $ Star ((Solo $!) . f)
+  in getSolo . runStar star
```
adamgundry, line 111: Should we move `set'` and `over'` to `Optics.Traversal` instead?

optics-core/optics-core.cabal
```diff
@@ -77,6 +77,7 @@ library
                , indexed-profunctors    >= 0.1        && <0.2
                , transformers           >= 0.5        && <0.7
                , indexed-traversable    >= 0.1        && <0.2
+               , OneTuple               >= 0.3        && <0.4
```
adamgundry, line 80: This looks like it might pull in quite a few extra dependencies to `optics-core`. Is that right, and can we avoid it?

optics-core/src/Optics/Setter.hs
```diff
@@ -128,10 +142,11 @@ set o = over o . const
 
 -- | Apply a setter, strictly.
 --
--- TODO DOC: what exactly is the strictness property?
---
+-- The new value will be forced if and only if the optic traverses at
+-- least one target. If forcing the new value is inexpensive, then it
+-- is cheaper to do so manually and use 'set'.
```
adamgundry, line 147: I wonder if we could easily test this, since we didn't before and apparently didn't get it right...

optics-core/src/Optics/Setter.hs
```diff
@@ -102,14 +102,28 @@ over o = \f -> runFunArrow $ getOptic (castOptic @A_Setter o) (FunArrow f)
 -- 'over' is used, because the first coordinate of a pair is never forced.
 --
 over'
```
adamgundry, line 104: The haddocks above here have another "TODO DOC" it would be great to nail along with this.

adamgundry (commented):
Thanks very much for contributing @treeowl! Sorry for the lag in responding. I'm not extremely familiar with the strictness stuff here, but the general idea seems plausible to me.

@arybczak would you be able to take a look, as you are probably more familiar with the relevant code?

optics-core/optics-core.cabal
adamgundry: Actually, perhaps this is true only for older GHC versions?

treeowl: It looks to me like the package is quite careful to avoid excessive dependencies when possible.

optics-core/src/Optics/Setter.hs
treeowl: I'm not sure what the intention is here. I'm just documenting what it actually does.

optics-core/src/Optics/Setter.hs
treeowl: They would be a better fit, probably; I was just trying to avoid breaking things unnecessarily.

optics-core/src/Optics/Setter.hs
treeowl: Or were you referring to testing strictness generally? `containers` does some of that; I haven't checked if there's traversal strictness testing, but if so that could be copied maybe.

optics-core/src/Optics/Setter.hs
treeowl: The intuition is that each target result is forced before producing the full result. More precisely, I believe

```haskell
over' p f xs = forceElements res `seq` res
  where
    forceElements = foldlOf' p (\_ !_ -> ()) ()
    res = over p f xs
```

optics-core/optics-core.cabal
arybczak: For psychological reasons I'd prefer us to not include this dependency.

The cost of not doing so is irrelevant - Solo can be copy pasted where `Identity'` was along with its Applicative instance.

treeowl: @arybczak Psychological reasons? Why put the compat code here when someone else has already written it there?

arybczak: > Psychological reasons? 

Yes, people don't like a lot of deps in "core" packages. Whether it's rational or not is irrelevant.

treeowl: Okay ... I'll copy whatever's needed over.

optics-core/src/Optics/Setter.hs
treeowl: That "law" only works with a non-type-changing traversal, but I think it's a good way to express the concept.

arybczak:
Superseded by #553.

PR #480, opened 2023-02-20
Add GHC-9.6.1 CI job

phadej:


phadej:
@adamgundry do you know what

```
src/Optics/SOP.hs:76:35: error: [GHC-18872]
    • Couldn't match representation of type: Optic
                                               A_Prism
                                               NoIx
                                               (NS (NP I) (x : xs))
                                               (NS (NP I) (x : xs))
                                               (NP I a6)
                                               (NP I a6)
                               with that of: Optic
                                               m0 ks0 (NS f1 (x1 : xs1)) (NS f1 (x1 : ys0)) a1 b0
```

might be related to?

adamgundry:
I haven't looked closely, but there have been some similar-smelling GHC issues recently: https://gitlab.haskell.org/ghc/ghc/-/issues/22924

Looks like it will be worth trying again with the RC.

phadej:
I tried with a GHC built from `ghc-9.6` branch yesterday, and the coercing works!

There's one test failing locally:

```
     traverseOf_ itraversed = traverseOf_ folded:                                   FAIL
        tests/Optics/Tests/Utils.hs:68:
        tests/Optics/Tests/Core.hs:99:21: lhs14 ==~ rhs14 passed.
        Use -p '/traverseOf_ itraversed = traverseOf_ folded/' to rerun this test only.
```

That test is expected to fail with GHC-9.2 and GHC-9.4, but seems to work again with GHC-9.6!

PR #481, opened 2023-02-21
Fix `Optics.Review.review` link in docs for `Optics.Iso` module

skykanin:
`Optics.Iso` documentation linked to `Optics.Getter.review` which doesn't exist, instead of `Optics.Review.review`.


arybczak:
Thanks!

PR #483, opened 2023-03-12
PoC: Add non-empty folds

phadej:
As `Foldable1` is in base-4.18 and there is a light compatibility package, we can do this somewhat easily.

This breaks symmetry of hierarchy,as there is no NeTraversal, but it's much less useful than non-empty folds (i.e. to semigroups).

This (softly) needs a `Foldable1WithIndex` to be added to `indexed-traversable`, which I'll do some time in the future (after GHC-9.6 updates are settled).

### DIFF
--- a/codegen/Subtypes.hs
+++ b/codegen/Subtypes.hs
@@ -40,6 +40,8 @@ data OpticKind
   |  A_Getter
   -- | Tag for an affine fold.
   |  An_AffineFold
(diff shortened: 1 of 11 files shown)


phadej:
Updated with a proof-of-concept of indexed folds too.

I'll leave this at this state until someone else asks for non-empty folds, but I'm probably going to release `indexed-traversable-0.1.3` in near future.

PR #485, opened 2023-03-24
Remove note about missing documentation for reexported modules

georgefst:
_As far as I can tell_, this is no longer relevant: Hackage now has documentation for every module.

Although I can't actually find anything recent in the Cabal/Haddock/Hackage trackers/histories indicating when this was fixed.


adamgundry:
Thanks! This has actually been fixed just for `optics` by manually uploading documentation that works around the problem, thanks to @phadej in #465. As far as I'm aware it remains an issue on Hackage in general.

PR #495, opened 2023-09-14
Support th-abstraction 0.6

Vekhir:
Builds and tests fine

### DIFF
--- a/optics-th/optics-th.cabal
+++ b/optics-th/optics-th.cabal
@@ -65,7 +65,7 @@ library
                , mtl                    >= 2.2.2     && <2.4
                , optics-core            >= 0.4.1     && <0.5
                , template-haskell       >= 2.12      && <2.21
-               , th-abstraction         >= 0.4       && <0.6
+               , th-abstraction         >= 0.4       && <0.7
                , transformers           >= 0.5       && <0.7
 
   exposed-modules: Optics.TH


adamgundry:
Thanks for the patch! I ran CI and it looks like there is a glitch in the doctests on GHC <= 9.2.8 (but this might well not be due to your change...).
```
src/Data/HashSet/Optics.hs:36: failure in expression `setOf folded ["hello","world"]'
expected: fromList ["hello","world"]
 but got: fromList ["world","hello"]
                     ^

Examples: 97  Tried: 96  Errors: 0  Failures: 1
```

phadej:
> but this might well not be due to your change

That's most likely a change in `hashable`, which doesn't guarantee hash values to stay stable, so `toList` of `HashSet` may list values in any order. After all, it's defined in `unordered-containers` package!

phadej:
FWIW, e.g. GHC-9.6.2 job picked up  `th-abstraction-0.5.0.0 (lib) (requires download & build)` (check *dump install plan* build step).

Testing locally, the reason is:

```
[__4] rejecting: th-abstraction-0.6.0.0 (conflict: generics-sop =>
th-abstraction>=0.4 && <0.6)
```

... so this PR seems to be merged prematurely.

adamgundry:
I've tested locally that `optics-th` builds and passes tests with `th-abstraction-0.6.0.0` (using `allow-newer`). Is that enough or should we revert?

`template-haskell-optics` seems to be hit by an API change in `th-abstraction-0.6.0.0`, so it doesn't build.

phadej:
> I've tested locally that 

Please mention these, that would prevent confusion.

> should we revert?

No need to, as long as someone remembers to make `template-haskell-optics` compatible as well.

adamgundry:
Got it.

I've bumped the bounds of https://hackage.haskell.org/package/optics-th-0.4.1 to permit `th-abstraction-0.6`.

PR #508, opened 2024-05-13
Add GHC-9.10.1 to CI, allow containers-0.7

phadej:
`template-haskell-optics` needs code changes, so I left it out.


phadej:
made revisions for optics-core, optics-extra optics-th and optics, i.e. packages with containers bounds.

PR #512, opened 2025-01-04
Add GHC-9.12 to CI

phadej:


erikd:
This is a GHC bug and I have raised an issue: https://gitlab.haskell.org/ghc/ghc/-/issues/25631

phadej:
@adamgundry @arybczak, in https://gitlab.haskell.org/ghc/ghc/-/issues/25631#note_605017 @nomeata suggests to disable failing tests. I don't like that (I'm afraid the test will stay disabled indefinitely) but if you both are OK with that, then we should do it to unblock this.

arybczak:
FYI I updated template-haskell bounds of optics-th on hackage.

phadej:
> FYI I updated template-haskell bounds of optics-th on hackage.

Please commit them too then, so they don't get lost.

arybczak:
Let's just merge this PR. I'm completely fine with disabling the test, considering that there is nothing going on in the GHC ticket.

adamgundry:
I agree, let's just disable the failing test on 9.12 until https://github.com/nomeata/inspection-testing/issues/82 is resolved. Thanks for pushing this through!

adamgundry:
I've opened https://gitlab.haskell.org/ghc/ghc/-/issues/25883 to track this upstream.

adamgundry:
I've got a proposed fix in https://gitlab.haskell.org/ghc/ghc/-/commit/02f58d00f54a4f1737013c34731f91779efebc70. Compiling with that patch only two test failures remain (plus one test starts passing again). I think these are just trivial differences in the order of let-bindings so we can presumably skip the affected tests on 9.12?

```
      iover (itraversed..itraversed) = iover (imapped..imapped):                     FAIL
        tests/Optics/Tests/Utils.hs:65:
        tests/Optics/Tests/Core.hs:85:20: lhs11 === rhs11 failed:
        LHS:
            lhs11
              = ... let {
                      $dFunctorWithIndex
                        = $p1TraversableWithIndex $dTraversableWithIndex_a2dc3 } in
                    let { f1 = imap $dFunctorWithIndex } in
                    imap
                      ($p1TraversableWithIndex $dTraversableWithIndex_a2dc2)
                      (\ i1 ->
                         let { g = f1 (\ i1 -> eta_B0 (i1, i1)) } in
                         \ x ->
                           case x of {
                             Left y -> Left y;
                             Right y -> Right (g y)
                           })
        RHS:
            rhs11
              = ... let {
                      $dFunctorWithIndex
                        = $p1TraversableWithIndex $dTraversableWithIndex_a2dc3 } in
                    imap
                      ($p1TraversableWithIndex $dTraversableWithIndex_a2dc2)
                      (\ i1 ->
                         let { g = imap $dFunctorWithIndex (\ i1 -> eta_B0 (i1, i1)) } in
                         \ x ->
                           case x of {
                             Left y -> Left y;
                             Right y -> Right (g y)
                           })
```

```
      iset (itraversed..) = iset (imapped..):                                        FAIL
        tests/Optics/Tests/Utils.hs:65:
        tests/Optics/Tests/Core.hs:120:20: lhs16 === rhs16 failed:
        LHS:
            lhs16
              = ... let {
                      $dFunctorWithIndex
                        = $p1TraversableWithIndex $dTraversableWithIndex_a2cXY } in
                    let { f1 = imap $dFunctorWithIndex } in
                    imap
                      ($p1TraversableWithIndex $dTraversableWithIndex_a2cXX)
                      (\ i1 ->
                         let { g = f1 (\ i1 _ -> eta_B0 (i1, i1)) } in
                         \ x ->
                           case x of {
                             Left y -> Left y;
                             Right y -> Right (g y)
                           })
        RHS:
            rhs16
              = ... let {
                      $dFunctorWithIndex
                        = $p1TraversableWithIndex $dTraversableWithIndex_a2cXY } in
                    imap
                      ($p1TraversableWithIndex $dTraversableWithIndex_a2cXX)
                      (\ i1 ->
                         let { g = imap $dFunctorWithIndex (\ i1 _ -> eta_B0 (i1, i1)) } in
                         \ x ->
                           case x of {
                             Left y -> Left y;
                             Right y -> Right (g y)
                           })
```

phadej:
> Unfortunately it looks like half of the test suite will get disabled over time due to these small differences,

It's a indeed a reality of `inspection-testing`. As our experience with real-world use-cases grows, the `inspection-testing` should be extended with larger equivalences. If we see that the only difference is order of lets, and it indeed doesn't matter, we should be able to make `inspection-testing` ignore that.

Unfortunately, it seems that `inspection-testing` is in a development limbo, but still, IMO the least we should do is to open issues to record the limitations.

nomeata:
> If we see that the only difference is order of lets, and it indeed doesn't matter, we should be able to make inspection-testing ignore that.

I think it already tries to do that by default:
https://github.com/nomeata/inspection-testing/blob/5afb1ee9ee949c74e5a1ee8bf5301cf36ec099a9/src/Test/Inspection/Core.hs#L375

That said, someone interested in helping/driving/taking over maintenance would be very welcome to do so!

phadej:
@nomeata 

> I think it already tries to do that by default:

As my latter comment says, it's not just order of lets, and I'm not sure the difference is trivial. So IMHO the `inspection-testring` failure here is legit, the code is different enough.

phadej:
The GHC issue @adamgundry made is resolved, so we are waiting for GHC-9.12.3. I'll check that the fix indeed makes tests pass. (Not sure whether the patch has backpoted to 9.12 branch though).

adamgundry:
Sounds good, thanks @phadej!

PR #516, opened 2025-03-24
Add GHC 9.10 support to `template-haskell-optics`

ozkutuk:


ozkutuk:
Looks like GHC 9.10 checks had been explicitly disabled in `haskell-ci` config: https://github.com/well-typed/optics/blob/1aa8a58a9d528acffdeac4a92d5b437175879a39/.github/workflows/haskell-ci.yml#L245

I have now re-enabled them by removing the handwritten version checks.

arybczak:
> I have now re-enabled them by removing the handwritten version checks

They are not handwritten, they're autogenerated by `haskell-ci`. Please regenerate the workflow file by running `haskell-ci regenerate` in the root repo dir (you'll need haskell-ci from git afaik).

ozkutuk:
> Please regenerate the workflow file by running `haskell-ci regenerate` in the root repo dir

I have now regenerated the config using haskell-ci head. There seems to be some GHCJS-related changes in the diff though; should I cherrypick the relevant sections or is it fine as is?

arybczak:
It's fine, looks like ghcjs support was dropped.

arybczak:
Thanks :)

PR #541, opened 2026-01-30
Implement `imaximumOf` and `iminimumOf`

### LINKED ISSUE #536: Idea: `imaximumOf` and `iminimumOf`
Recently I needed a way to find the index of the maximum element in a list, and I was surprised that there was not an indexed counterpart to `maximumOf`. I came up with the following, which I thought could also be generally useful to have in `optics`:

```haskell
newtype IxMin i a = IxMin { getIxMin :: Maybe (a, i) }
  deriving (Semigroup, Monoid) via Maybe (Min (a, i))

iminimumOf
  :: (Is k A_Fold, is `HasSingleIndex` i, Ord a, Ord i)
  => Optic' k is s a -> s -> Maybe (i, a)
iminimumOf o = fmap swap . getIxMin . ifoldMapOf o (curry (IxMin . Just . swap))

newtype IxMax i a = IxMax { getIxMax :: Maybe (a, i) }
  deriving (Semigroup, Monoid) via Maybe (Max (a, Down i))

imaximumOf
  :: (Is k A_Fold, is `HasSingleIndex` i, Ord a, Ord i)
  => Optic' k is s a -> s -> Maybe (i, a)
imaximumOf o = fmap swap . getIxMax . ifoldMapOf o (curry (IxMax . Just . swap))
```

I wrote them such that if there exists multiple maximum elements, the one with the "lesser" index is returned:

```haskell
>>> iminimumOf ifolded "calamari"
Just (1,'a')
```

Is it desirable to have these in the library? If so, I can prepare a PR for them.

ozkutuk:
Resolves #536.

### DIFF
--- a/optics-core/src/Optics/IxFold.hs
+++ b/optics-core/src/Optics/IxFold.hs
@@ -36,6 +36,8 @@ module Optics.IxFold
   , inoneOf
   , ifindOf
   , ifindMOf
+  , imaximumOf
+  , iminimumOf
 
   -- * Combinators
   , ipre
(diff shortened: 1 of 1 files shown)


ozkutuk:
@arybczak Rebased onto latest master, and all checks pass now.

arybczak:
This looked like it leaks space (definitely the `min` bit), so I ran the review with Claude Fable. Here's the feedback:

EDIT: I consolidated the first version with my doubts about taking min of the index and the imaximumOf disparity, here's the final feedback:

**1. The `EQ` branch can return an (index, value) pair that never occurred in the fold**

  ```haskell
  EQ -> IxMax (min i1 i2, a1)
  ```

  This takes the min of the indices but unconditionally the *left* value. For a fold
  whose traversal order isn't monotone in the index (e.g. `backwards_ ifolded`, or a
  hand-rolled `ifoldVL`), a tie with `i2 < i1` returns `(i2, a1)` — index from one
  element, value from another. This is observable whenever `compare` returns `EQ` for
  distinguishable values. Whatever tie-breaking we settle on, whole elements must be
  kept — never a mix of components.

  **2. Tie-breaking semantics and the `Ord i` constraint**

  The non-indexed `minimumOf` uses `min`, which keeps the *first* of equal elements,
  while `maximumOf` uses `max`, which keeps the *last*. I'd like the law

  ```haskell
  fmap snd . iminimumOf o ≡ minimumOf o   -- and same for maximum
  ```

  to hold exactly, even for distinguishable-but-`EQ` values. That means: on a tie,
  `iminimumOf` keeps the earlier element and `imaximumOf` the later one. This also
  drops the `Ord i` constraint entirely (it exists only to serve the least-index
  tie-break), which is the simpler signature — and since the constraint hasn't shipped
  yet, now is the time: removing it later is fine, but adding it later would be
  breaking. For ordinary monotone folds this coincides with "least index for the
  minimum, greatest index for the maximum" anyway.

  **3. Space-leak-prone implementation, inconsistent with the non-indexed versions**

  `maximumOf`/`minimumOf` in `Optics.Fold` deliberately use `foldlOf'` with strict
  accumulation, and their haddocks explicitly warn that the `foldMapOf`-with-a-semigroup
  formulation "has lazier semantics but could leak memory". The new functions use
  exactly that formulation: `ifoldMapOf` over `Maybe (IxMin i a)`, where `Maybe`'s `<>`
  builds a thunk chain.

  `ifoldlOf'` is already available in this module, so together with points 1 and 2 the
  implementation simplifies to (no helper newtypes needed):

  ```haskell
  iminimumOf
    :: (Is k A_Fold, is `HasSingleIndex` i, Ord a)
    => Optic' k is s a -> s -> Maybe (i, a)
  iminimumOf o = ifoldlOf' o mf Nothing where
    mf i acc a = case acc of
      Just (_, a') | compare a a' /= LT -> acc
      _                                 -> Just (i, a)

  imaximumOf
    :: (Is k A_Fold, is `HasSingleIndex` i, Ord a)
    => Optic' k is s a -> s -> Maybe (i, a)
  imaximumOf o = ifoldlOf' o mf Nothing where
    mf i acc a = case acc of
      Just (_, a') | compare a a' == LT -> acc
      _                                 -> Just (i, a)
  ```

  (`ifoldlOf'` forces the accumulator to WHNF at each step, and since ties now keep an
  existing element rather than computing anything, nothing accumulates.)

  **4. Missing `INLINE` pragmas**

  Every other exported combinator in this file has one (cf. the 0.4.1.1 release, which
  existed solely to add missing `INLINE` pragmas). `imaximumOf` and `iminimumOf` should
  have them too.

  **5. Wrong `@since` annotation**

  Both functions say `@since 0.4.3`, but the last release was 0.4.2 and master is
  already at `optics-core-0.5`. These should be `@since 0.5`.

  **6. Missing changelog entry**

  The 0.5 section of `optics-core/CHANGELOG.md` doesn't mention these functions; the
  entry should also be mirrored in the `optics` changelog.

  Minor doc-parity point: the non-indexed versions note that on a valid
  `Iso`/`Lens`/`Getter` they always return `Just`, plus the strictness remark — worth
  adding to the indexed haddocks as well. The tie-breaking behavior from point 2 should
  also be documented.

arybczak:
Would be also good to add `*ByOf` variants for completeness.

arybczak:
Superseded by #548.

PR #547, opened 2026-07-22
Drop support for GHC older than 9.2

### LINKED ISSUE #519: Composable `ghc*failure`.
With more GHC versions we end up with predicates looking like

```haskell
ghc92and94and9122failure :: Result -> IO ()
#if (__GLASGOW_HASKELL__ >= 902 && __GLASGOW_HASKELL__ <= 904) || GHC_9122
ghc92and94and9122failure = assertFailure'
#else
ghc92and94and9122failure = assertSuccess
#endif
```

I suggest that we refactor these so we'd rather write tests as

```haskell
assertSucceessExcept (ghc92 || ghc94 || ghc9122) $ $(lhsFoo ==~ rhsFoo)
```

### LINKED ISSUE #532: Drop support for older GHCs
With GHC-9.14 we'll support 13 major GHCs. That's quite many CI jobs. I think keeping it at around 10 (~5 years worth of GHCs) should more than enough.

arybczak:
Fixes #519, fixes #532, drops massive amounts of CPP.

I think dropping <9.2 is reasonable, it's still 4.5 years worth of GHC releases.

I'll leave it for a few days in case anyone wanna protest.


adamgundry:
Thanks very much for picking this up @arybczak! I'm in support of dropping GHC versions prior to 9.2.

phadej:
Issue of #519 is not fixed by  this, AFAICT. (The particular example is, but the core problem isn't).

arybczak:
@phadej how come? Conditional asserts now look like this: `assertFailureIf (ghcVer <= (9,4))`, so you can compose and add version ranges without any new CPP.

ozkutuk:
Qualified imports of `Data.Semigroup` could also be dropped now I think, maybe something to consider for a future clean-up.

phadej:
@arybczak I misread the code.  It was unclear how solving "Drop support for GHC older than 9.2" solved #547, but this patch had independent changes: dropping support for older GHCs didn't require solving #547, nor vice versa (being helpful is not a hard requirement). I'm personally *not* a fan of such bundling, even if it creates slightly more work. I think it's important that independent changes are kept independent, especially in a project with multiple contributors, because it's then easier to keep track of changes and staying up to date.

